PR #49, opened 2011-09-01
Allow specifying a scope function

s0undt3ch:
Allow specifying a scope function.

### DIFF
--- a/flaskext/sqlalchemy.py
+++ b/flaskext/sqlalchemy.py
@@ -620,6 +620,8 @@ class SQLAlchemy(object):
         """
         if options is None:
             options = {}
+        if scopefunc is None:
+            scopefunc = _request_ctx_stack.__ident_func__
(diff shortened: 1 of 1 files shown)


flaskext/sqlalchemy.py
```diff
@@ -620,6 +620,8 @@ def create_scoped_session(self, options=None, scopefunc=None):
         """
         if options is None:
             options = {}
+        if scopefunc is None:
+            scopefunc = _request_ctx_stack.__ident_func__
```
fantix: This function does not return ident of current request (or current app in latest code base), instead it returns the ident of current thread. (this function is the built-in `_thread.get_ident` by default)
@davidism See https://github.com/pallets/werkzeug/blob/b3637ff4039200cc60c1a65ade422127127722a2/werkzeug/local.py#L121

Issue #62, opened 2011-11-20
Allow specifying a different Model as a base


s0undt3ch:
Right now, there's no simple way to provide a base Model object because everything is done on the `make_declarative_base` function, the options is to override and redo what's done on that function. Could we provide a different Model when creating the SQLAlchemy object? or provide the model as a class attribute(since I'm already subclassing it)?

PR #72, opened 2012-02-28
Allow user-specified Model base classes

jfinkels:
This is in response to issue #62. It allows a user to specify a base class for models (which should be a subclass of ``flaskext.sqlalchemy.Model``) in the constructor of ``flaskext.sqlalchemy.SQLALchemy``. **Note**: this does not work with the ``SQLAlchemy.init_app(app)`` idiom unless the model class is specified in the constructor of ``SQLAlchemy``.


ThiefMaster:
:+1: - this would still be increadibly useful and avoid nasty monkeypatching!

davidism:
#328

PR #78, opened 2012-05-25
Fixed bug in models_commited behavior.

gyllstromk:
If multiple instances from different models added to session and committed, any PK overlap among models will cause collisions and only one instance per PK gets sent to models_committed.

Updated test suite to capture new behavior.


immunda:
Fixed in master, soon to be released in 2.1

PR #79, opened 2012-05-28
Modification of storing "sqlalchemy_queries" (debug) from deprecated Con...

djanku:
...nectionProxy interface to recomended event listening "before_cursor_execute" and "after_cursos_execute".

Corrected dot "flask.ext.sqlalchemy" in source file header.


davidism:
1625f5343f230bf6749ad2701b2f71aa01131a6e

PR #85, opened 2012-06-15
Allow optional message on 404 from db query

robcowie:
`description` arg on `.get_or_404()` and `.first_or_404()` is passed to abort, which in turn passes it to the NotFound exception.

For example, the following will return a 404 response with 'Unknown User' in it's body.

```python
User.query.get_or_404('Unknown user')
```

### DIFF
--- a/flask_sqlalchemy.py
+++ b/flask_sqlalchemy.py
@@ -351,22 +351,22 @@ class BaseQuery(orm.Query):
     standard query as well.
     """
 
-    def get_or_404(self, ident):
+    def get_or_404(self, ident, description=None):
         """Like :meth:`get` but aborts with 404 if not found instead of
         returning `None`.
         """
         rv = self.get(ident)
         if rv is None:
-            abort(404)
+            abort(404, description=description)
         return rv
 
-    def first_or_404(self):
+    def first_or_404(self, description=None):
         """Like :meth:`first` but aborts with 404 if not found instead of
         returning `None`.
         """
         rv = self.first()
         if rv is None:
-            abort(404)
+            abort(404, description=description)
         return rv
 
     def paginate(self, page, per_page=20, error_out=True):


immunda:
Rather a specific use case, I don't think this belongs in the lib.

robcowie:
That was a long wait for rejection 😁

immunda:
Sorry man, I hope this gave you closure. 🙏

davidism:
@immunda I kind of like that this mirrors how `abort(404)` works.  It's not a big change, but the alternative to this is overriding `Query`, which is not trivial right now, or catching the 404 and re-raising.

immunda:
Maybe I'm biased as I've never really made use of the `or_404` helpers :wink:

A more flexible `get_or_abort`, which was basically just a proxy to the actual `abort` would be handy. For API compatibility, we could retain the `or_404` methods as shortcuts to `get_or_abort(404, ...)`

So, how about this plan @davidism?

* Merge this as is (@robcowie could you rebase this to `master` and ideally make a note in `CHANGES`?)
* Bring in `get_or_abort` which is a pretty transparent proxy for `abort` as part of v3
* Also eat `get_or_abort` dogfood for `or_404` methods in v3

davidism:
That sounds good.  I'm not sure how "correct" an error besides 404 is for a missing resource, but I'm sure I've seen questions regarding that use case before.

immunda:
One that springs to mind would be querying for a user by some credential and returning a `401` or `403` when that fails.

bastianh:
I don't think that query is the right place for get / get_or_404 in the first place. 
I suggest creating a function in your class for that (using a mixin) like CRUDmixin.

davidism:
@bastianh No way, this is way too common and convenient to move somewhere else.  Also, `get` is a method on SQLAlchemy's Query, so if you don't think it shouldn't be there you should file a bug with SQLAlchemy.  Adding extra methods to Query is a standard practice, `get_or_404` just mirrors what's already there.

bastianh:
I don't suggest removing the feature .. I just don't think that this feature should get extended even more to handle every possibility with different models.

If a model need special treatment (because it's a user model for example) it is quite easy to add such a function to the model. There is no need to alter the library or override query.

I think it's important for the library to do things sqlalchemy way.. like encouraging using a raw declarative_base of sqlalchemy and making it compatible with other sqlalchemy addon libraries instead of adding new functions that makes it even more incompatible. It should not alter the way you use sqlalchemy or make it easier or whatever...

```
class AbortMixin(object):
    @classmethod
    def get_or_abort(cls, id, code=404, reason=None):
        obj = db.session.query(cls).get(id)
        if not obj:
            abort(code, reason)
        return obj
```

This simple snippet will allow what the user wants.. you can simply add it to every model if you want and it will also work with plain sqlalchemy without bloating the library with a function for every possible usecase.

davidism:
Adding methods to query *is* the "SQLAlchemy way".  Adding that mixin to every model is way less convenient and also requires every developer to know how and remember to add it every time.  I'd love to chat about this more, but I don't think this is the right place for that.

bastianh:
I don't think this is a chat... it's a discussion about wether it is necessary or reasonable to add specific functionality to this library.

I think that the django example is supporting my case ... they don't have added that shortcut to the query class.. you have to import it seperatly from django.shortcuts.

Adding the functionality as a mixin was just an example.  You could also adding it to your basemodel or make it a function you import from a shortcut class like django

>  The point is not to "add a method to cover every use case" (we're not adding any methods in this pr)
no .. but there were also comments in this pr where this is heading in the future (v3)

the base functionality is there and I don't think it is necessary to cover every case. last week I needed a get_or_404 function for a rest api that returns a json object with status code 404 so I wrote this function and added it to my project. Of couse I could have made a pull request for a `get_or_404(id, as_json)` but  I think that is something to specific to add to to library.

davidism:
It's great that you came up with a solution for your JSON api.  Given that you would have had to do that whether this function existed on `query` or not, I'm not clear what your objection is.  The only thing your suggestion does is create more work for everyone who's *not* writing a custom 404 JSON response.

justanr:
I think a more general `get_or` and `get_or_raise` pair would be more approriate, but that's just me. I tend to avoid `get_or_404` just because I don't want the ORM interfering with my routing and error handling. However, once #319 is hashed out, everyone can be satisfied.

PR #89, opened 2012-07-23
"Fix for AttributeError: 'SessionMaker' object has no attribute '_model_changes' when using fixture to load fixtures"

jpanganiban:
Hi,

I encountered the same error as this (scenario is also stated in the link): http://flask.pocoo.org/mailinglist/archive/2011/3/6/flask-sqlalchemy-and-fixtures/.

Thanks


jquacinella:
+1 for new release

Issue #90, opened 2012-07-25
before_models_committed signal never sent


kalymero:
It seems the before_models_committed is never sent.

Looking in depth in this piece of code:

class _SignalTrackingMapperExtension(MapperExtension):

    def after_delete(self, mapper, connection, instance):
        return self._record(mapper, instance, 'delete')

    def after_insert(self, mapper, connection, instance):
        return self._record(mapper, instance, 'insert')

    def after_update(self, mapper, connection, instance):
        return self._record(mapper, instance, 'update')

    def _record(self, mapper, model, operation):
        pk = tuple(mapper.primary_key_from_instance(model))
        orm.object_session(model)._model_changes[pk] = (model, operation)
        return EXT_CONTINUE

class _SignallingSessionExtension(SessionExtension):

    def before_commit(self, session):
        d = session._model_changes
        if d:
            before_models_committed.send(session.app, changes=d.values())
        return EXT_CONTINUE

    def after_commit(self, session):
        d = session._model_changes
        if d:
            models_committed.send(session.app, changes=d.values())
            d.clear()
        return EXT_CONTINUE

    def after_rollback(self, session):
        session._model_changes.clear()
        return EXT_CONTINUE

the problem seems like this:

* In the case of after_commit the flow i can see is:
  - _record method called
  - session._model_changes filled
  - after_commit called
  - d is filled with _model_changes
  - the signal is sent

* In the case of before_commit, the flow is:
  - before_commit is called
  - d in empty so no signal is sent
  - _record is called and _model_changes filled
  - but it's to late... 

but i can't figure out how this happen

immunda:
Fixed in master, soon to be released in 2.1

PR #95, opened 2012-08-08
Fix before_models_committed signal

insom:
As per #90 - I'm seeing the same thing; the signal never fires because `d` is always empty when it does. I originally added `before_{insert,update,delete}` event handlers, but it looks like `before_commit` fires before even they do.

The approach on this pull request is to introspect the session object and see what is queued to be committed. I've included tests done by @DazWorrall which originally failed but now pass and I believe cover what the documentation says this signal should do.


immunda:
This is better addressed in #150

Issue #130, opened 2013-05-31
Using ShardedSession with flask-sqlalchemy


malicustommade:
Hi,

I'm looking into using a custom session, the SQLAlchemy ShardedSession, but I don't think there's a simple way to do that so far.

DasIch (Daniel Neuhäuser) in #pocoo mentioned that you'd probably do this by modifying `create_scoped_session()` in the flask-sqlalchemy `SQLAlchemy` class.

This function uses `_SignallingSession`. I'm looking at the `_SignallingSession` here: https://github.com/mitsuhiko/flask-sqlalchemy/blob/master/flask_sqlalchemy.py#L156 and it doesn't seem to do too much other than manage the binds. The other class is here: https://github.com/zzzeek/sqlalchemy/blob/master/lib/sqlalchemy/ext/horizontal_shard.py

I'd love to submit a patch to add sharding support but there's a few ways we could handle this:

1) Don't implement this at all but instead allow people to pass in their own session class, and provide a decorator that wraps their custom `get_bind()` to do the right thing, which they have to use. This is more flexible but less batteries-included.
2) Merge these two together: Write our own `ShardedSession`/`SignallingSession` hybrid to handle the binds properly. But now we have to maintain it, which is probably a miniscule amount of work really.
3) Do both: use 1 with 2

Also, designwise, I'm not sure if the binds in each paradigm mean the same thing. I guess it could be that you can specify binds as you do now, but you can also specify that certain shards are comprised of certain binds. So then if you specify a bind, it'll query only that. If you specify a shard, it'll know to query all the sub-binds. Does this sound reasonable?

Anyone have any guidance?

rsyring:
This should get taken care of by: https://github.com/pallets/flask-sqlalchemy/issues/438

PR #138, opened 2013-06-24
Add support for custom Session class and Query class

malicustommade:
... and consolidate all places where BaseQuery is hardcoded, to use SQLAlchemy.query_class. Basically this patch is to increase configurability for the end user without mucking with internals much.

There was an instance where setting query_class in the Model didn't change the Query objects on relationship / relation / dynamic_loader as used in _include_sqlalchemy(). At first I thought this was a bug but then after reading the SQLAlchemy class docstring it looks like it's by design? Anyway, I "fixed" it and I added a test case for that, but let me know and I can revert that.

This patch is part of a larger project to integrate sharding support (or make a flask.ext.sharding), as in #130, so it may or may not make sense to integrate this in at this point, but I'd love to hear comments from @DasIch and @mitsuhiko to steer me in the right direction.

### DIFF
--- a/flask_sqlalchemy.py
+++ b/flask_sqlalchemy.py
@@ -633,14 +633,8 @@ class SQLAlchemy(object):
 
     def __init__(self, app=None,
                  use_native_unicode=True,
-                 session_options=None,
-                 session_class=None):
+                 session_options=None):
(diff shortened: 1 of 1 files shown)


balsagoth:
Hi, this feature was merged?

PR #146, opened 2013-08-03
Added 'SQLALCHEMY_ENCODING' option

lordkrandel:
Allows SQLAlchemy engine to deal with different encodings


rsyring:
Thanks for contributing and sorry it's taken so long to get you a reply. This functionality will be supported by #684 which I hope to get merged and released soon.

lordkrandel:
You'll never know how this makes me proud; even if it's a small contribution and a late reply, it's so nice to see you care. Thank you!

PR #149, opened 2013-08-15
added ability to do runtime configuration of database schema

meantheory:
Preface -> not sure I know what I am doing as I am new to sqlalchemy and flask.

As far as I have been able to understand from the docs and playing around with everything is there is no way to do runtime configuration of the schema. There is a lot to understand here so my apologies if this is already possible via other means.

My specific use case is that I want to have the same set of table's in multiple schemas inside the postgres database. 

Example: client1.users and client2.users

This is easy to do in SQLAlchemy but I couldn't figure out a way to make the flask sqlalchemy extension perform this. My goal with my hack was to enable this extension to do runtime configuration. 

This appears to already be possible in mysql as the URI configuration seems to allow you to specify a schema but I could find no corollary for postgres.

My apologies for any bad code. Just wanted to get the idea out there.


davidism:
#255 allows custom metadata

PR #117, opened 2013-02-19
Support for before_models_flushed and models_flushed signals

bkabrda:
The first commit includes a fix for breakage of sqlalchemy hooks done by mistake in mitsuhiko/flask-sqlalchemy@cf40caec1c48fc2188597ca4e6ddfb21b770d9d7

The second commit introduces support for before_models_flushed and models_flushed signals, closing mitsuhiko/flask-sqlalchemy#9

PR #150, opened 2013-08-23
fix before_models_committed to actually fire with stuff

tonymillion:
As highlighted by @insom before_models_committed will not fire because it has no changes.

Also see https://github.com/mitsuhiko/flask-sqlalchemy/pull/117
and https://github.com/mitsuhiko/flask-sqlalchemy/pull/95

Its plainly obvious this is broken (even the simplest of test cases shows it) When will any of these fixed be pulled into main???

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -179,11 +179,15 @@ class _SessionSignalEvents(object):
 
     @staticmethod
     def session_signal_before_commit(session):
-        if not isinstance(session, SignallingSession):
-            return
-        d = session._model_changes
-        if d:
-            before_models_committed.send(session.app, changes=d.values())
+        if session._new or session._deleted or session.identity_map:
+            s = []
+            for new in session._new:
+                s.append((new.obj(), 'insert'))
+            for delete in session._deleted:
+                s.append((delete.obj(), 'delete'))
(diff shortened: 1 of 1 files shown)


immunda:
FWIW: looking into this now, but this approach doesn't actually work as it's not targeting the signals correctly.

tonymillion:
FYI: I don't actually use any of this any more, I use the SQLAlchemy events instead.

nickretallack:
Perhaps it should be removed from the documentation?

davidism:
The fix in this patch is not correct.  The session only tracks new, dirty, and deleted objects since the last *flush*.  By the time the *commit* occurs these collections may be empty or incomplete.  Also, there is no reason to use the "private" attributes rather than `new`, `dirty`, and `deleted`.  Finally, if `_model_changes` is going to be ignored, the events that populate it shouldn't be registered in the first place, and the `models_committed` signal should be changed similarly.

These signals and events are pretty clunky compared to using SQLAlchemy's events and history tracking directly.  I agree they should be deprecated and eventually removed.

immunda:
Yep, the more I look into these Flask-SQLAlchemy signals, the less I think they should be included.

We should have some docs about using SQLAlchemy signals, remove the documentation for the existing signals and then schedule the existing signals for removal in 3.0.

PR #156, opened 2013-10-01
Signal won't be fired on operating multiple models having same PK

weboo:
When insert or update multiple tables in one transaction, only one signal will be fired if those tables has same primary key.
To solve this problem, I added the tablename to _model_changes key.


immunda:
Fixed in master, soon to be released in 2.1

PR #161, opened 2013-10-15
Fix: consider base classes while determining table names (without reduce)

wienand:
I am not sure if creating a new pull request is the way to go. But I did not find a way to update the old closed one #124.

I changed the reduce magic to a more straight iteration over the base classes. 

Example use case: Standard mixin which defines the table names

```python
class StandardMixin(object):
  @declared_attr
  def __tablename__(cls):
    return cls.__name__.lower()
```


davidism:
This is addressed by #248. @immunda

immunda:
:+1:

PR #168, opened 2013-11-23
Add support for binds parameter to session

jpvanhal:
This allows one to pass ``binds`` parameter to the signaling session. My use case for this is that I want to override the binds in tests with connections wrapped in a transaction so that I can rollback the transaction in test teardown, which will make the tests run faster.


rawrgulmuffins:
Any update on getting this merge in?

immunda:
Resolved by the merge of #249

PR #169, opened 2013-11-27
Fix a bug where options were not passed correctly to the create_session ...

oargon:
The way it's written does not allow to pass parameters for a ad-hoc session which is different from the global configuration and will fail when SQLAlchemy will try to call the create_session function.
For example, this will not work without the fix:
```python
app = Flask(__name__)
db = SQLAlchemy(app)
with app.app_context():
    s = db.session(expire_on_commit=False)
    s.add(SomeModel)
    s.commit()
```


immunda:
Duped in #219

PR #183, opened 2014-03-17
Convert to relative imports

AlJohri:
Using "flask.ext.sqlalchemy" within the import creates problems for programs like pyinstaller, py2app, py2exe, or cx-freeze.

These programs freeze the executables and linked python packages in a codebase by recursively checking import statements throughout all dependencies.

Since flask uses redirection from flask.ext.sqlalchmey to flask_sqlalchemy, the import statements are not understood unless one explicitly adds a "hidden_import" of flask_sqlalchemy.

Upon adding this hidden_import, the code works fine, only if relative imports are used within the codebase.

Otherwise, individual hidden_imports must be added for each import, like flask.ext.sqlalchmey._compat.


davidism:
839f50ca24bdac66b750554117e55af6f336d742

PR #186, opened 2014-03-31
Honor SQLALCHEMY_RECORD_QUERIES even in debug mode. Fix #181

### LINKED ISSUE #181: Memory leak in DEBUG mode
Flask-SQLAlchemy record all SQL queries in debug mode. I agree it is useful to debug but it also consume all my memory in long running processes. E.g. a celery task importing a lot of data to my database.

I suggest a hard limit to record only last N queries.

Also, if I just set `SQLALCHEMY_RECORD_QUERIES=False`, it is ignored because it always record in debug mode.

iurisilvio:


rsyring:
Thank you for your contribution, sorry this reply is so long in coming.  This dups #320.  Keeping that PR b/c it also updates the docs.

PR #189, opened 2014-04-16
Scoped Session returns instance object

iamjarret:
This updates the SQLAlchemy object so that the sessions it generates, when used to add an object
to the database, returns an instance of the object added.

Use case:
t = db.session.query(Topbooks).filter('user'==current_user).first() or db.session.add(Topbooks(user=current_user))


immunda:
This is a little odd and doesn't belong here.

P.s. Why not just make your own `get_or_create` method?

```
def get_or_create_book(user):
    book = db.session.query(Topbooks).filter('user'==user).first()
    if book is None:
        book = Topbooks(user=user)
        db.session.add(book)
    return book
```

PR #196, opened 2014-05-07
Debug queries with bindkey

seyhbold:
When I have a flask_sqlalchemy setup with multiple binds, I would like to see the bind key in the `_DebugQueryTuple`. Without this key tools like `FlaskDebugToolbar`are unable to work with the query, i.e. view select results or explain plan.


immunda:
I think overriding components would be fine for your specific use-case. Unfortunately this PR is also full of cruft.

PR #208, opened 2014-07-06
support 'echo_pool' option when creating the engine.

yonil:
I would like to add support for turning on the 'echo_pool' option when creating the engine.

http://docs.sqlalchemy.org/en/rel_0_9/core/engines.html#sqlalchemy.create_engine.params.echo_pool


immunda:
Mind giving this a rebase?

rsyring:
Thanks for contributing and sorry it's taken so long to get you a reply.  This will be supported by https://github.com/pallets/flask-sqlalchemy/pull/684 which I hope to get merged and released soon.

PR #215, opened 2014-08-19
Allow mysql to use NullPool

cbronazc:
If pool_size is set to None, then NullPool should be used.

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -777,6 +777,12 @@ class SQLAlchemy(object):
             if info.drivername != 'mysql+gaerdbms':
                 options.setdefault('pool_size', 10)
                 options.setdefault('pool_recycle', 7200)
+            pool_size = options.get('pool_size')
+            if not pool_size:
+                from sqlalchemy.pool import NullPool
+                options['poolclass'] = NullPool
+                options.pop('pool_size', None)
+
         elif info.drivername == 'sqlite':
             pool_size = options.get('pool_size')
             detected_in_memory = False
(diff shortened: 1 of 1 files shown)


thoutenbos:
You can override it without patching the original class if you want until a definitive fix is merged

```python
from flask_sqlalchemy import SQLAlchemy as SQLAlchemyBase
from sqlalchemy.pool import NullPool

class SQLAlchemy(SQLAlchemyBase):
  def apply_driver_hacks(self, app, info, options):
    super(SQLAlchemy, self).apply_driver_hacks(app, info, options)
    options['poolclass'] = NullPool
    options.pop('pool_size', None)

db = SQLAlchemy()
```

rsyring:
https://github.com/pallets/flask-sqlalchemy/pull/684 will make it easier to set engine options.

@davidism what do you think about this PR?  It's essentially doing what we already do for SQLite when pool_size is None or 0.  So it seems like it would be ok.  But I have two questions:

1) If we are now doing this for SQLite and MySQL, what about other drivers.  Why should these two receive special treatment?
2) If we want to do this, do we need to wait for 3.0?  It technically changes behavior, by changing the pool class that would be used, but it seems like it would have no practical difference on usage since a NullPool or StaticPool with poolsize=0 would have the same behavior?

rsyring:
Actually, now that I'm looking at the PR a bit closer, this is wrong.  If pool_size is not set, then FSA would not pass in that value to `create_engine()`.  That should result in the usage of SA's [QueuePool](https://docs.sqlalchemy.org/en/latest/core/pooling.html#sqlalchemy.pool.QueuePool) which has a default `pool_size=5`.

As best I can tell, the acceptance of this PR would result in any default MySQL usage going from using a connection pool of 5 to disabling connection pooling through the use of `NullPool`.

Feel free to post back here if you think I misunderstanding.  Thanks.

PR #219, opened 2014-08-31
Fixed a bug where the options were not being passed correctly

oargon:
This is the same as pull request 169 that didn't get in for some reason.
I also added a test so that this will be accepted now before version 2.0.

See also: https://github.com/mitsuhiko/flask-sqlalchemy/pull/169


davidism:
This is possible (and was already possible).  `db.session` is a `scoped_session`, which allows you to call it with extra arguments.  If the session has already been used in this scope, [the call will fail unless you call `db.session.remove()` first](http://docs.sqlalchemy.org/en/rel_1_0/orm/contextual.html#sqlalchemy.orm.scoping.scoped_session.__call__).

PR #235, opened 2014-10-28
added SQLALCHEMY_DATABASE_OPTIONS to config settings 

unixfreak0037:
This allows users to customize their call to create_engine as they need to.  Specifically, this allows me to use SSL for MySQL.  I have it merge into what already exists.

<pre>class DevelopmentConfig(Config):
    SQLALCHEMY_DATABASE_URI = 'mysql+mysqldb://user:password@127.0.0.1/database'
    SQLALCHEMY_DATABASE_OPTIONS = {
        'connect_args': {
            'ssl' : {
                'key': 'ssl/mysql/client-key.pem',
                'cert': 'ssl/mysql/client-cert.pem',
                #'cert': 'ssl/mysql/ca-cert.pem'
            }
        }
    }</pre>

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -492,6 +492,9 @@ class _EngineConnector(object):
                 return self._engine
             info = make_url(uri)
             options = {'convert_unicode': True}
+            # allow user to specify additional parameters to the create_engine function
+            if 'SQLALCHEMY_DATABASE_OPTIONS' in self._app.config:
+                options.update(self._app.config['SQLALCHEMY_DATABASE_OPTIONS'])
             self._sa.apply_pool_defaults(self._app, options)
             self._sa.apply_driver_hacks(self._app, info, options)
             if echo:


DocOtak:
I would like to see this (or something like it) merged in. I think this would also solve/close pull requests #208 and #215 in a nice generic solution.

If the reason this hasn't been merged in yet is made clear, I'll help resolve it.

banxi1988:
PR like this should be merged in. I trying hard to find way to specify some `create_engine` options such as `json_serializer` but failed.

lovmat:
I need this too (also for setting json_serializer). This PR also solves:
https://github.com/mitsuhiko/flask-sqlalchemy/issues/450

(I personally prefer this solution over the one suggested in the above ticket)

cburmeister:
You can pass SSL arguments as query parameters in the `SQLALCHEMY_DATABASE_URI`:
```
SQLALCHEMY_DATABASE_URI = 'mysql://user:password@host/db?ssl_key=&ssl_ca=&ssl_cert='
```

alexanderkoumis:
Hi, any movement on this? Need this patch badly to pass a different 'connect_args': {'timeout': N} value to create_engine.

cburmeister:
@alexanderkoumis `timeout` is pretty ambiguous. Which engine are you using?

Perhaps you're looking for `SQLALCHEMY_POOL_TIMEOUT`?

alexanderkoumis:
I am using sqlite, I need to bump the value of `timeout` used here: https://docs.python.org/2/library/sqlite3.html#sqlite3.connect (also see http://stackoverflow.com/questions/15065037/how-to-increase-connection-timeout-using-sqlalchemy-with-sqlite-in-python).

Reading through the source it appears that `SQLALCHEMY_POOL_TIMEOUT` corresponds with the 'pool_timeout' arg used with QueuePool (http://docs.sqlalchemy.org/en/latest/core/pooling.html#sqlalchemy.pool.QueuePool):

> pool_timeout=30 – number of seconds to wait before giving up on getting a connection from the pool. **This is only used with QueuePool**.

This page also mentions:

> In the case of SQLite, the SingletonThreadPool or NullPool are selected by the dialect

So it does not appear that setting the `SQLALCHEMY_POOL_TIMEOUT` value will increase the correct timeout. I am using a hack akin to https://github.com/mitsuhiko/flask-sqlalchemy/pull/215#issuecomment-233603952 for now but it would be great to have a natively supported solution. Thanks!

jfach:
+1

I'm pretty much in the exact same scenario as the OP here. Has anyone been able to get around this?

rsyring:
Thank you for your interest and sorry it's taken so long to get a reply and solution for this issue.  https://github.com/pallets/flask-sqlalchemy/pull/684 should get merged soon to address this need.

PR #236, opened 2014-10-30
Test case for before_models_committed

zhangcheng:


immunda:
I'm reverting this because after digging deeper there's way more to it than this test.

Issue #204, opened 2014-06-10
SQLAlchemy.Model subclasses do not honor __tablename__ override


ghost:
I have a custom BaseModel class that derives from `SQLAchemy(app).Model` which sets up a `__tablename__` via declared_attrs. However that is not honored by flask-sqlalchemy because `_BoundDeclarativeMeta` does not look at the bases to see if `__tablename__` is already provided.

My BaseModel class:

```
class BaseModel(DB.Model):
     __abstract__ = True

    @declared_attr
    def __tablename__(cls):
        namespace = cls.__module__.replace('.models', '')
        name = '%s_%s' % (namespace, cls.__name__.lower())
        return name[:-5] if name.endswith('model') else name
```

This will merely namespace the tables and remove suffixes. 

What I would expect is that if I have a model class in module `bar` defined as:

```
class TestModel(BaseModel):
    foo = DB.Column(DB.Integer)
```

its table name would be `bar_test`.

If we look up bases in `_BoundDeclarativeMeta` to see if `__tablename__` is overridden and use that it appears to work as desired:

```
class _BoundDeclarativeMeta(DeclarativeMeta):

    def __new__(cls, name, bases, d):
        tablename = d.get('__tablename__')
        if not tablename:
            for base in bases:
                if hasattr(base, '__tablename__'):
                    tablename = getattr(base, '__tablename__')
                    break
```

Issue #198, opened 2014-05-13
Model hierarhy inherit parent table name


tbicr:
Next code example:

```
from datetime import datetime

from flask import Flask
from flask.ext.sqlalchemy import SQLAlchemy
from sqlalchemy import Column, Integer, DateTime, String


app = Flask(__name__)
app.config['SQLALCHEMY_DATABASE_URI'] = 'sqlite://'
db = SQLAlchemy(app)


class BaseModel(db.Model):

    __abstract__ = True

    id = Column(Integer, primary_key=True)
    create_datetime = Column(DateTime, default=datetime.now)
    update_datetime = Column(DateTime, default=datetime.now, onupdate=datetime.now)


class TestModel(BaseModel):

    item = Column(String(64))


with app.app_context():
    db.create_all()
    db.session.commit()
    for table in db.metadata.tables:
        print(table)
```

will return: `base_model`, but expected `test_model`.

Issue #83, opened 2012-06-08
Broken support of __abstract__ mark


ivirabyan:
```
class User(db.Model):
    __abstract__ = True
    id = Column(Integer, primary_key=True)

class Manager(User):
    name = Column(String)
```

This code must declare only one table `manager`, with two fields `id` and `name`. But Flask-SQLAlchemy doesn't generate `__tablename__` for class Manager, because it doesn't have a primary key field. So, accessing `Manager.__tablename__` results in `user`, which obviously not what I want.

PR #248, opened 2014-12-16
Support generating tablenames in more cases.

davidism:
Extends naming rules to apply to base classes of models.  This supports joined table and single table inheritance, inheriting from mixins and abstract models, and _some_ complex inheritance combinations.

Resolves the following issues:
- mitsuhiko/flask-sqlalchemy#83
- mitsuhiko/flask-sqlalchemy#198
- mitsuhiko/flask-sqlalchemy#204

I am almost 100% confident that this does not break existing behavior; however, the original behavior was not well documented or tested.  This does seem to cover all "normal" cases that I've run into, and I've added tests for these cases.  I did not test concrete table inheritance in its various forms as I am not familiar enough with it.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -46,3 +46,4 @@ Patches and Suggestions
 - Daniel Lepage
 - Ignacy Sokołowski
 - Steven Harms
+- David Lord @davidism
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,14 @@ Changelog
 Here you can see the full list of changes between each Flask-SQLAlchemy
 release.
 
+Version 2.1
+-----------
+
+In development
+
+- Table names are automatically generated in more cases, including
+  subclassing mixins and abstract models.
+
 Version 2.0
 -----------
 
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -504,24 +504,51 @@ class _EngineConnector(object):
             return rv
 
 
-def _defines_primary_key(d):
-    """Figures out if the given dictionary defines a primary key column."""
-    return any(v.primary_key for k, v in iteritems(d)
-               if isinstance(v, sqlalchemy.Column))
+def _should_set_tablename(bases, d):
+    """Check what values are set by a class and its bases to determine if a
+    tablename should be automatically generated.
+
+    The class and its bases are checked in order of precedence: the class
+    itself then each base in the order they were given at class definition.
+
+    Abstract classes do not generate a tablename, although they may have set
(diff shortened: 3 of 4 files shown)


seth-p:
Does this work for concrete table inheritance, e.g. something like this?
```
class X(AbstractConcreteBase, Base):
    idx = sa.Column(sa.Integer, primary_key=True)

    @declared_attr
    def __mapper_args__(cls):
        if cls.__name__ == 'X':
            return {}
        else:
            return {'polymorphic_identity': cls.__name__,
                    'concrete': True}

class Y1(X):
    float_field = sa.Column(sa.Float)

class Y2(X):
    date_field = sa.Column(sa.Float)
```

it seems like it would set ``__tablename__`` for ``X``, but then not for ``Y1`` and ``Y2``, which would simply inherit ``__tablename__ = 'x'``.

I'm afraid I don't actually use Flask, but am looking at ``_should_set_tablename`` to address a similar issue in https://github.com/dgilland/alchy. So apologies for not testing this myself.

davidism:
@seth-p like I say in the first message, I do not use concrete inheritance, so I didn't try to add checks for it.  There were no checks before, and all the tests still pass.  I can try to add some tests for this example you've provided, soon™.

seth-p:
Thanks. FWIW, the following is the logic for dealing with ``__abstract__`` in sqlalchemy\ext\declarative\base.py. It seems quite subtle, and I can't say I follow it. I suspect that to handle concrete table inheritance properly ``_should_set_tablename()`` may need to implement similar logic.
```
def _as_declarative(cls, classname, dict_):

    ...

    for base in cls.__mro__:
        _is_declarative_inherits = hasattr(base, '_decl_class_registry')

        ...

        if '__abstract__' in base.__dict__ and base.__abstract__:
            if (base is cls or
                    (base in cls.__bases__ and not _is_declarative_inherits)):
                return
```

[More generally, what is the downside to autogenerating ``__tablename__`` for any class that doesn't have ``__tablename__`` and ``__table__`` in its ``__dict__`` (regardless of whether it's abstract)? Are there ever situations where one wants a class to inherit ``__tablename__`` unmodified from its parent? Or would that break the ``@declared_attr`` usage described in #204?]

seth-p:
FWIW, in https://github.com/dgilland/alchy/pull/20/files I implemented a sightly modified version of _should_set_table_name() that supports concrete inheritance, and included tests of all inheritance types. The handling of  '__global_table_args__', '__local_table_args__', and 'query' is specific to Alchy, but otherwise I would think the code should work as is in Flash-SQLAlchemy.

PR #252, opened 2015-01-09
Remove "is it ready"

svenstaro:
This is a remnant of a time when the version number wasn't quite as encouraging.


svenstaro:
How can this possibly fail travis? Must have been another commit at fault.

immunda:
This didn't break the tests. Taking on board removal of "is it ready", but I'd rather it became "what is the current version?"

PR #254, opened 2015-01-09
Add additional note about tox

svenstaro:
Because currently it's available but not documented.


svenstaro:
What do you think about this?

immunda:
Looks good.

PR #61, opened 2011-11-15
Allow specification of external metadata object to be used by the SQLAlchemy object. (issue #60)

nathan-rice:
I use my model definitions outside my web application, so being able to decouple them is nice.  Also, I dislike circular references; I would prefer everything go model definition module -> web application module.

PR #255, opened 2015-01-10
Add custom meta data parameter

### LINKED ISSUE #60: No way to use a pre-existing metadata object with flask-sqlalchemy
I would prefer not to couple my modeling code to my web application.  My preferred technique is to create a metadata in the same module as my model definitions, passing that later to table definitions.  This lets me use the same module definitions in external scripts without a lot of overhead.

svenstaro:
Fixes #60, #61.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -11,6 +11,7 @@ In development
 
 - Table names are automatically generated in more cases, including
   subclassing mixins and abstract models.
+- Allow using a custom MetaData object.
 
 Version 2.0
 -----------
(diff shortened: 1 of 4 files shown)


davidism:
It's clear we need to be able to override *everything* that is passed to `make_declarative_base`.  A more thorough way to do this would to be allowed to pass in a `base_kwargs` dictionary, which will be `**` expanded.

svenstaro:
Any more discussion on this? Is that the last word on it?

immunda:
@davidism is right, but let's get this in for 2.1. A more general solution can be developed 3.0 which will accept `metadata` as part of `kwargs`.

Issue #237, opened 2014-10-30
No before_models_committed signal emitted


zhangcheng:
I created a [test case](https://github.com/mitsuhiko/flask-sqlalchemy/pull/236) for it, but I have no clear idea how to fix it.

immunda:
Thanks, closing as it duplicates #90

Issue #162, opened 2013-10-22
AttributeError: 'Session' object has no attribute '_model_changes'


yang:
I'm getting

```
AttributeError: 'Session' object has no attribute '_model_changes'
```

when trying to use a plain sqlalchemy Session elsewhere in unrelated code that is not flask-sqlalchemy-aware, e.g. simply calling:

```
sqlalchemy.orm.Session(sqlalchemy.create_engine('sqlite:///')).commit()
```

immunda:
Fixed in master, soon to be released in 2.1

Issue #160, opened 2013-10-13
test_sqlalchemy.SignallingTestCase.test_model_signals() fails with Python 3


Arfrever:
`test_sqlalchemy.SignallingTestCase.test_model_signals()` fails with Python 3.

```
$ python3.3 setup.py test
...
======================================================================
FAIL: test_model_signals (test_sqlalchemy.SignallingTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/flask-sqlalchemy/test_sqlalchemy.py", line 142, in test_model_signals
    self.db.session.commit()
  File "/usr/lib64/python3.3/site-packages/sqlalchemy/orm/scoping.py", line 149, in do
    return getattr(self.registry(), name)(*args, **kwargs)
  File "/usr/lib64/python3.3/site-packages/sqlalchemy/orm/session.py", line 721, in commit
    self.transaction.commit()
  File "/usr/lib64/python3.3/site-packages/sqlalchemy/orm/session.py", line 361, in commit
    self.session.dispatch.after_commit(self.session)
  File "/usr/lib64/python3.3/site-packages/sqlalchemy/event.py", line 370, in __call__
    fn(*args, **kw)
  File "/tmp/flask-sqlalchemy/flask_sqlalchemy/__init__.py", line 194, in session_signal_after_commit
    models_committed.send(session.app, changes=d.values())
  File "/usr/lib64/python3.3/site-packages/blinker/base.py", line 267, in send
    for receiver in self.receivers_for(sender)]
  File "/usr/lib64/python3.3/site-packages/blinker/base.py", line 267, in <listcomp>
    for receiver in self.receivers_for(sender)]
  File "/tmp/flask-sqlalchemy/test_sqlalchemy.py", line 135, in committed
    self.assertTrue(isinstance(changes, list))
AssertionError: False is not true

----------------------------------------------------------------------
Ran 21 tests in 0.779s

FAILED (failures=1)
```

Type of `changes` with Python 3 is `dict_values`, not `list`.
Fix:

```
--- test_sqlalchemy.py
+++ test_sqlalchemy.py
@@ -132,7 +132,7 @@
     def test_model_signals(self):
         recorded = []
         def committed(sender, changes):
-            self.assertTrue(isinstance(changes, list))
+            self.assertTrue(isinstance(changes, type({}.values())))
             recorded.extend(changes)
         with sqlalchemy.models_committed.connected_to(committed,
                                                       sender=self.app):
```

Issue #144, opened 2013-07-22
Make signalling part of SignallingSession optional


rslinckx:
I don't use any of the signalling features of the SignallingSession, and I especially need to avoid the _model_changes dict being filled during a transaction because I have a use case where i need to perform a massive transaction before committing and the changes tend to fill the memory until the transaction is committed.

Could a config option be implemented to avoid tracking model changes at all ?

PR #256, opened 2015-01-11
Fix and deprecate change tracking

davidism:
Use `before_flush` event to track changes.  This removes the need for mapper events.  It also makes the `before_models_committed` event work, addressing #90, #95, #150, and #237.  Instances are identified using SQLAlchemy's `identity_key` if possible rather than just primary key, to guard against separate models with the same keys, addressing #78 and #156.

Only register modification tracking events if the configuration enables them.  Previously, the events would always be registered but wouldn't do anything.  This reduces the overhead of handling each callback when not needed.  The events are registered on specific sessions, not the top-level session, and registration creates the `_model_changes` property, so there is no need to guard against this not existing anymore.  This better addresses #89, #109, #135, #144, and #162.

Discussion in #150 concluded that the signals should be deprecated. The default config is changed to `None` to indicate "not set".  This enables the events and issues a warning that the events will be disabled by default in the future.  Explicitly setting the config to `True` silences the warning, as that will be the normal way to enable the events in the future.  Based on this, #117 should be closed since adding *more* signals doesn't make sense.

The test for `before_models_committed` (#236) was always failing because the variable in closure couln't be set.  This is fixed by using a namespace and setting an attribute there.  There is also an issue with the type of `changes` (#160), this is fixed by always returning a list (which matches the behavior of the other signal).

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -51,10 +51,10 @@ import pkg_resources
 try:
     release = pkg_resources.get_distribution('Flask-SQLAlchemy').version
 except pkg_resources.DistributionNotFound:
-    print 'To build the documentation, The distribution information of'
-    print 'Flask-SQLAlchemy has to be available.  Either install the package'
-    print 'into your development environment or run "setup.py develop"'
-    print 'to setup the metadata.  A virtualenv is recommended!'
+    print('To build the documentation, The distribution information of')
+    print('Flask-SQLAlchemy has to be available.  Either install the package')
+    print('into your development environment or run "setup.py develop"')
+    print('to setup the metadata.  A virtualenv is recommended!')
     sys.exit(1)
 del pkg_resources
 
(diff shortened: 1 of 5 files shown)


immunda:
@davidism We keep missing each other on IRC (-:

Keen to have this in, but I think it might be worth slating for 3.0 as it's a bit more of a drastic change.

davidism:
@immunda I try to check the logs but they're not the most searchable.

This patch is mostly about fixing broken behavior.  All that will change (for the end-developer) by merging this now is that the two signals will behave correctly (and actually work at all).  A warning will be printed once stating that this will be disabled.  There are no backwards incompatible changes yet though.

I was thinking that version 3 would change the behavior to disabled by default, but I think this patch is appropriate for version 2.  It's pretty drastic, but that's all behind the scenes, all existing tests still pass.

immunda:
Let's have this in for 2.1

Issue #213, opened 2014-08-02
"does not have a __table__ or __tablename__ specified" but the real issue is lack of ids


endolith:
I am completely new to SQL and in my first attempts I get the error:

> InvalidRequestError: Class <class 'flask_app.Votes'> does not have a __table__ or __tablename__ specified and does not inherit from an existing table-mapped class.

which is very strange because another tables works fine despite not having a __table__ or __tablename__

Then I find out, as described in http://stackoverflow.com/a/14595840/125507 that this is just a confusing error message and the real problem is lack of an `id` column, so please make the error message more helpful

immunda:
Addressed in #258

cgomezmendez:
I got this exact thing happening on version Flask-SQLAlchemy==2.3.0 and was pretty confuse.

davidism:
@cgomezmendez please open a new issue with a reproducible example.

PR #258, opened 2015-01-11
document automatic tablename

davidism:
Mention that a tablename will be generated if the model sets a primary key and doesn't otherwise set a tablename.  Related to #248.  Addresses #213.


davidism:
Failing tests are not related.

PR #265, opened 2015-02-21
Allows for SQLAlchemy Query object to be paginated through Flask-SQLAlchemy

neurosnap:
http://stackoverflow.com/a/15734565/1713216

I ran into this issue recently and thought adding a quick fix in Flask-SQLAlchemy could be useful to others.

What this PR does is allows an SQLAlchemy Query object: `db.session.query()` to use all the same functionality as `BaseQuery.paginate()`.

This very simply moves the primary functionality from `BaseQuery.paginate()` into its own function and now `BaseQuery.paginate()` calls that function.

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -301,7 +301,7 @@ def get_debug_queries():
 
 
 class Pagination(object):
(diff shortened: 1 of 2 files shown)


davidism:
#328 means this works now

PR #271, opened 2015-03-11
Change import to reflect deprecated ext format

lepture:
Use flask_sqlalchemy instead of flask.ext.sqlalchemy on module name. https://github.com/mitsuhiko/flask-sqlalchemy/pull/270

https://github.com/mitsuhiko/flask/issues/1135

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1,7 +1,7 @@
 API
 ---
 
-.. module:: flask.ext.sqlalchemy
+.. module:: flask_sqlalchemy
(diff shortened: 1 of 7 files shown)


keyanp:
I left this change out of #270 because I think it breaks doc generation as it did for https://github.com/thadeusb/flask-cache/pull/111

davidism:
I know this was the earlier one, but I saw #321 first, which I merged, so closing this now.

PR #320, opened 2015-06-16
Have the SQLALCHEMY_RECORD_QUERIES setting take precedence over DEBUG

csdev:
This allows query recording to be disabled regardless of the other application settings

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -28,9 +28,11 @@ A list of configuration keys currently understood by the extension:
                                    be useful for debugging.
 ``SQLALCHEMY_RECORD_QUERIES``      Can be used to explicitly disable or
                                    enable query recording.  Query recording
-                                   automatically happens in debug or testing
-                                   mode.  See :func:`get_debug_queries` for
-                                   more information.
+                                   is turned on by default in debug or
+                                   testing mode, but that behavior can be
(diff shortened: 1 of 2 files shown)


bastianh:
You know that documentation says that queries are always recorded in debug mode? maybe the pull request should also change the documentation ( if it gets accepted )

immunda:
Right on @bastianh, I'm always cautious to change existing and documented behaviour like this.

Have you come a cropper with the current precedence @csdev?

csdev:
@immunda that is definitely a valid concern. This is a change in application behavior that should probably wait for a major release. The reason I am proposing it is because it would be nice to disable query recording even though debug mode is on. In one of my projects, we have some database tasks that run a large set of queries. We have found that query recording can use up an excessive amount of memory.

@bastianh I will take a look at the docs later and add those changes to the pull request.

jeffwidman:
Dupe of #186 (although I prefer this version as it also updates the docs)

rsyring:
Thank you for your contribution and sorry this reply is so long in coming.  Setting the milestone for this PR to 3.0 since it changes observable behavior.

davidism:
This is fixed in #1087. However, I'm considering removing the automatic behavior and always requiring the config to be set.

PR #321, opened 2015-06-24
Replace flask.ext.sqlalchemy with flask_sqlalchemy

ThiefMaster:


jeffwidman:
:+1:

davidism:
Can't believe I missed this, since I've been commenting about it everywhere else.  Thanks.

immunda:
Glad to have this in, but was it confirmed not to break things as mentioned in #271 ?

davidism:
I just verified that docs build correctly on Python 2.7.10 and 3.5.0.  All cross references work and the API is generated.

immunda:
Perfect, ta.

PR #328, opened 2015-08-09
Implement custom query and model class options

justanr:
Implement proposal in #319 Specify Custom Query Class.

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -737,6 +737,7 @@ class SQLAlchemy(object):
             session_options = {}
 
         session_options.setdefault('scopefunc', connection_stack.__ident_func__)
+        session_options.setdefault('query_cls', query_class)
         self.use_native_unicode = use_native_unicode
         self.session = self.create_scoped_session(session_options)
         self.Model = self.make_declarative_base(model_class, metadata)
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -740,10 +740,8 @@ class SQLAlchemy(object):
         session_options.setdefault('query_cls', query_class)
         self.use_native_unicode = use_native_unicode
         self.session = self.create_scoped_session(session_options)
-        self.Model = self.make_declarative_base(model_class, metadata)
-        # ugly hack, is there a better way?
-        self.Model.query_class = query_class
         self.Query = query_class
+        self.Model = self.make_declarative_base(model_class, metadata)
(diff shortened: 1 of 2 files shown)


justanr:
If everything is alright with this, I'll also modify the documentation.

justanr:
Argh. I just realized this pull doesn't set the query class on the session options. It'll be tomorrow before I can fix that.

flask_sqlalchemy/__init__.py
```diff
@@ -730,19 +730,23 @@ class User(db.Model):
...
 
         session_options.setdefault('scopefunc', connection_stack.__ident_func__)
+        session_options.setdefault('query_cls', query_class)
         self.use_native_unicode = use_native_unicode
         self.session = self.create_scoped_session(session_options)
-        self.Model = self.make_declarative_base(metadata)
-        self.Query = BaseQuery
+        self.Model = self.make_declarative_base(model_class, metadata)
+        # ugly hack, is there a better way?
+        self.Model.query_class = query_class
```
justanr: I'm not happy with this line. I'd considered moving it to `make_declarative_base` but I'm not 100%. Should we respect the query class on the model, leave this as is/move to `make_declarative_base`, add a toggle for replacing the query class or something else?

davidism: Everything dealing with the Model should be in one place.  Just do something like `if getattr(Model, 'query_class') is None: Model.query_class = query_class` to avoid overriding it.

davidism:
The issue here is that `relationship` won't behave completely as expected.  If I have a relationship to `MyModel` which has a custom `query_class` (custom as in specific to that model), the relationship will still use the generic `query_class` passed in to `_set_default_query_class`.

justanr:
@davidism You can specify a query class on the relationship with the `query_class` keyword. `_wrap_with_default_query_class` (that's a mouthful) uses `setdefault` which respects existing values.

If you're asking if we can introspect and pull that custom class off, I'm sure its possible but it'd be much more involved.

flask_sqlalchemy/__init__.py
justanr: I added a check to respect the existing `query_class` attribute on models, and this potential replacement was moved into `SQLAlchemy.make_declarative_base`. However, I'm wondering if this'll cause confusion with some people. Providing a model base with your desired `query_class` will work though.

justanr:
Any updates on this merge?

immunda:
LGTM, caveats from @davidism noted.

PR #342, opened 2015-10-27
Fixed the issue with pool_size with sqllite

srikalyan:
We use sqllite for unit testing out code we observed the following error
with flask_sqlalchmey 2.1

TypeError: Invalid argument(s) 'pool_size' sent to create_engine(), using
configuration SQLiteDialect_pysqlite/StaticPool/Engine.  Please check that
the keyword arguments are appropriate for this combination of components.

After investigating a bit, I've discovered that we were using wrong pooling
mechanism for sqllite in-memory dbs (StaticPool). You can get more details
http://docs.sqlalchemy.org/en/latest/core/pooling.html

Basically, for sqllite there only 2 pooling mechanisms i.e.,
1. SingletonThreadPool (in memory)
2. NullPool (file based)

So I have switched pooling mechanism to use SingletonThreadPool instead of
StaticPool to fix the issue.

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -861,8 +861,8 @@ class SQLAlchemy(object):
             # which is fail.  Let the user know that
             if info.database in (None, '', ':memory:'):
                 detected_in_memory = True
-                from sqlalchemy.pool import StaticPool
-                options['poolclass'] = StaticPool
+                from sqlalchemy.pool import SingletonThreadPool
+                options['poolclass'] = SingletonThreadPool
                 if 'connect_args' not in options:
                     options['connect_args'] = {}
                 options['connect_args']['check_same_thread'] = False


srikalyan:
Please let me know if you have questions or concerns. All our builds are broken due to 2.1 release and we were hoping if you can patch 2.1 release with this commit and release it as soon as possible.

Thanks.

immunda:
Thanks for the PR. Interestingly, this contradicts other SQLAlchemy advice regarding using a memory database with multiple threads http://docs.sqlalchemy.org/en/latest/dialects/sqlite.html#using-a-memory-database-in-multiple-threads and http://docs.sqlalchemy.org/en/latest/core/pooling.html#sqlalchemy.pool.SingletonThreadPool

There was a change since 2.0 to 2.1 in this commit which introduced the default configuration of `StaticPool` as the `poolclass` for SQLite `:memory:` connections https://github.com/mitsuhiko/flask-sqlalchemy/commit/9b5bd9427fb3562926b3f9d9f9401ea6159a1fb2

Could you give me some more detail about which version of `SQLAlchemy` you're using?

srikalyan:
In the first link you have posted, it also says that 
"""
SingletonThreadPool is used by the SQLite dialect automatically when a memory-based database is used. See SQLite.
""""
The problem with StaticPool is that it's constructor does not take any pool_size argument 
https://github.com/zzzeek/sqlalchemy/blob/master/lib/sqlalchemy/pool.py. 

But `apply_pool_defaults` method in SQLAlchemy class adds the pool_size to the configuration (so pool size is present by default). Also, `apply_driver_hacks` method does the following 
""""
 if pool_size == 0:
                    raise RuntimeError('SQLite in memory database with an '
                                       'empty queue not possible due to data '
                                       'loss.')
""""
to ensure that pool_size is present for sqllite driver name and database being memory. This contradicts the StaticPool as it does not have pool_size argument in the constructor so during the create_engine call any configuration that is not poped out would be thrown as exception in this case it is pool_size with the following message

"""
TypeError: Invalid argument(s) 'pool_size' sent to create_engine(), using
configuration SQLiteDialect_pysqlite/StaticPool/Engine. Please check that
the keyword arguments are appropriate for this combination of components.
""""

BTW, I am using latest sqlalchemy.

immunda:
I've tried this out and I can only replicate when `SQLALCHEMY_POOL_SIZE` is explicitly set or is not `None`, can you please confirm that's not in your config anywhere?

immunda:
Test case:

```
from flask import Flask; from flask.ext.sqlalchemy import SQLAlchemy
app = Flask(__name__)
db = SQLAlchemy(app)
db.create_all()
app.config['SQLALCHEMY_POOL_SIZE'] = 5
db = SQLAlchemy(app)
db.create_all()
[...] TypeError: Invalid argument(s) 'pool_size' sent to create_engine(), using configuration SQLiteDialect_pysqlite/StaticPool/Engine.  Please check that the keyword arguments are appropriate for this combination of components.
```

srikalyan:
Aha, I see it now ignore my PR sorry for the false alarm.

immunda:
No problem. Glad this is working for you now, got a bit of a fright there ;)

PR #343, opened 2015-10-27
#336 OFFSET must not be negative

### LINKED ISSUE #336: OFFSET must not be negative
I don't know if this is already reported, but when we have zero results for a pagination object, and we force ?page=-1 then we will get the error OFFSET must not be negative.

http://stackoverflow.com/questions/32814217/offset-must-not-be-negative

groteworld:
Why:
It is not valid to have a negative offset or page. These changes allow
`query.paginate` to run with `error_out` on but still default to page 1
(or offset 0)

Closes #336

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -47,3 +47,4 @@ Patches and Suggestions
 - Ignacy Sokołowski
 - Steven Harms
 - David Lord @davidism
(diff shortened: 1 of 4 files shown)


davidism:
continued in #558

PR #353, opened 2015-11-23
Clarify table name override attribute

jcomo:
There is a place in the docs where it mentions overriding the table name, but doesn't explain how to do it. I added a clarification. There might be a better place for this or it may be better left out altogether (and kept to the SQLAlchemy docs)


immunda:
Much obliged.

Issue #275, opened 2015-04-02
Deprecate Flask-SQLAlchemy specific events


immunda:

immunda:
#364 fixes problems that led to this thinking, and since `SQLALCHEMY_TRACK_MODIFICATIONS` will be disabled by default in future, there's no benefit in breaking compatibility for those that still want to make use of `_SessionSignalEvents`.

jeffwidman:
Are there Flask-SQLAlchemy Signals that cover situations not covered by SQLAlchemy's built-in events? 

I get that you don't want to break compatibility--that makes sense. But if there's no advantage here other than backwards compatibility, wouldn't it still make sense to mark the Flask-SQLAlchemy signals as deprecated to discourage new apps from using them? No need to remove them right now, but still mark as deprecated.

PR #358, opened 2015-12-04
Add support for nested sessions (SAVEPOINTs) in modifications tracking

barakalon:
I love the signals for tracking modifications to models. However, they don't yet support [nested sessions](http://docs.sqlalchemy.org/en/latest/orm/session_transaction.html#using-savepoint).

Here is a proposal for adding support for this. It uses a stack of dictionaries instead of just a dictionary for tracking models that have been inserted/updated/deleted. When a nested session begins, a new dictionary is pushed to the stack. If that nested session is rolled back, the changes that happened during that session are removed. If that nested session is committed, the changes are merged into the next item.

Anywho, let me know what you think! This is my first attempt at a contribution to flask-sqlalchemy, so take it easy on me :grin:

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -22,7 +22,7 @@ from flask import _request_ctx_stack, abort, has_request_context, request
 from flask.signals import Namespace
 from operator import itemgetter
 from threading import Lock
-from sqlalchemy import orm, event, inspect
+from sqlalchemy import orm, event
 from sqlalchemy.orm.exc import UnmappedClassError
 from sqlalchemy.orm.session import Session as SessionBase
 from sqlalchemy.engine.url import make_url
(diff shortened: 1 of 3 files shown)


davidism:
Thanks, this looks interesting.  I'll try to take a look at it soon, just commenting to let you know I'm aware of it. :-)

barakalon:
Thanks @davidism !

I've been using this in production for about a month now and it's working fine.

immunda:
Whilst this change in the current context looks good, there are plans to deprecate Flask-SQLAlchemy specific signals (#275), as SQLAlchemy signals have come a long way since this library was created and some of the signals are quite flawed. This is also inline with plans to not track modifications by default as it adversely affects performance.

Happy to have input on these decisions, especially @davidism as he's input on this thread.

barakalon:
@immunda Just to make sure I'm on the same page, you mean SQLAlchemy's [events](http://docs.sqlalchemy.org/en/latest/orm/events.html), yes?

immunda:
Sorry, yes, `events`. Some Django parlance is hard to shake.

I also think you should actually ignore my previous comment regarding #275, since #364 now fixes problems that led to that decision.

barakalon:
Ah, right on. 

Well when tracking modifications is disabled by default, support for savepoints will still be a good thing.

davidism:
Why did you close this?

barakalon:
Ah, just because its been open for many months and I was doing a bit of house cleaning. I figured it was dead.

roganov:
Would be interesting to know how much this affects performance since per @immunda even the current naive implementation affects performance "adversely".

ViktoriyaSavkina:
@davidism any plans to merge this?

flask_sqlalchemy/__init__.py
```diff
@@ -178,67 +178,86 @@ class _SessionSignalEvents(object):
     @classmethod
     def register(cls, session):
         if not hasattr(session, '_model_changes'):
```
fantix: should be `_model_changes_stack` here?

flask_sqlalchemy/__init__.py
```diff
@@ -178,67 +178,86 @@ class _SessionSignalEvents(object):
...
         event.listen(session, 'before_flush', cls.record_ops)
         event.listen(session, 'before_commit', cls.record_ops)
         event.listen(session, 'before_commit', cls.before_commit)
         event.listen(session, 'after_commit', cls.after_commit)
-        event.listen(session, 'after_rollback', cls.after_rollback)
+        event.listen(session, 'after_soft_rollback', cls.after_rollback)
 
     @classmethod
     def unregister(cls, session):
         if hasattr(session, '_model_changes'):
```
fantix: ditto

ffleandro:
+1 for this

davidism:
#1087 changes modification tracking, so this will no longer be mergeable. I also have the impression that how SQLAlchemy tracks nested transactions has changed, but I'm not sure. That said, after considering it more, I don't want to add more complexity to the track modifications feature. If you need other behavior, you're better off using SQLAlchemy's events directly.

Issue #182, opened 2014-03-17
Listening for 'before_flush' throws sqlalchemy.exc.ArgumentError


acowlikeobject:
Registering a listener for 'before_commit' (and 'before_flush') like so, where:
```python
db = SQLAlchemy(app)
...
event.listen(db.session, 'before_commit', before_commit)
```

Throws the following exception:
```python
    @event.listens_for(db.session, "before_flush")
  File "/home/environments/python3/lib/python3.3/site-packages/sqlalchemy/event/api.py", line 68, in decorate
    listen(target, identifier, fn, *args, **kw)
  File "/home/environments/python3/lib/python3.3/site-packages/sqlalchemy/event/api.py", line 49, in listen
    _event_key(target, identifier, fn).listen(*args, **kw)
  File "/home/environments/python3/lib/python3.3/site-packages/sqlalchemy/event/api.py", line 22, in _event_key
    tgt = evt_cls._accept_with(target)
  File "/home/environments/python3/lib/python3.3/site-packages/sqlalchemy/orm/events.py", line 1142, in _accept_with
    "Session event listen on a scoped_session "
sqlalchemy.exc.ArgumentError: Session event listen on a scoped_session requires that its creation callable is associated with the Session class.
```

Per the comment in this thread from Mike Bayer, it seems the issue is with the factory being passed to scoped_session not being a sessionmaker():
http://stackoverflow.com/questions/21322158/event-listener-on-scoped-session

Stepping through the code, it barfs in this method in line 1132 of sqlalchemy/orm/events.py:
```python
    @classmethod
    def _accept_with(cls, target):
        if isinstance(target, scoped_session):

            target = target.session_factory
            if not isinstance(target, sessionmaker) and \
                (
                    not isinstance(target, type) or
                    not issubclass(target, Session)
                ):
                raise exc.ArgumentError(
                            "Session event listen on a scoped_session "
                            "requires that its creation callable "
                            "is associated with the Session class.")
```

`db.session.session_factory` is of type `functools.partial`, which angers SQLAlchemy.

kbussell:
You can work around this by registering the event handler on the underlying session class instead of on the created session object:

    event.listen(SignallingSession, 'before_commit', before_commit)

nehresma:
@immunda bump.  I'm running into this as well.

opyate:
Same here.

davidism:
[...]

Issue #147, opened 2013-08-12
Use sessionmaker as the session factory and provide session-derived class


malthe:
Because ``sqlalchemy.orm.events.SessionEvents`` requires the "creation callable" of a  ``scoped_session`` to be "associated with the Session class.".

Issue #64, opened 2011-12-01
use of partial() in create_scoped_session prevents SQLAlchemy.session from being used with SQL Alchemy event listeners


nathan-rice:
SQL Alchemy events are registered by calling the listen function with a class, an event name and a function.  By returning a partial, you make it difficult to attach listeners.  I suppose you could work around this issue by calling type(db.session()), however a decent amount of debugging is required to figure out why things don't work in the first place.

A simple solution is to create a singalling_session_maker function that takes the same arguments as signalling_session, then define _SignallingSession inside this function with a reduced __init__ function, and return it.  Then you can just call that factory inside of scoped_session().

PR #364, opened 2015-12-14
Fix db.session so that events work

davidism:
`create_scoped_session` was using `partial` as the factory to `scoped_session`.  SQLAlchemy doesn't recognize this when listening to events, so it raises an exception.  Creating a `sessionmaker` in `create_session` fixes this.

    sqlalchemy.exc.ArgumentError: Session event listen on a scoped_session requires that its creation callable is associated with the Session class.

Also changed `create_scoped_session` to include the default `scopefunc` and `query_cls`, since it would be pretty much pointless to manually use this function without those provided.

#182, #64, #147, #148

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -8,7 +8,7 @@
     :copyright: (c) 2014 by Armin Ronacher, Daniel Neuhäuser.
     :license: BSD, see LICENSE for more details.
 """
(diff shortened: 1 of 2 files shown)


jeffwidman:
:+1:

immunda:
:clap:

jeffwidman:
Any chance of a Pypi release that includes this commit?

WhatsCS:
I too would like this to be pushed.

PR #366, opened 2015-12-18
Fix self.app not being set on init_app

sc4reful:
I create a SQLAlchemy instance and import it into my application's main file from my models file. I then call init_app and was getting errors when I tried to perform actions with this instance and realized SQLAlchemy.app was never set when calling init_app on it.


sc4reful:
Realized this is intentional behavior. Sorry.

PR #373, opened 2016-01-29
Support binds on abstract models

singingwolfboy:
Flask-SQLAlchemy is currently unable to handle defining a `__bind_key__` attribute on an abstract model. This change fixes that, and it includes an automated test.

### DIFF
--- a/test_sqlalchemy.py
+++ b/test_sqlalchemy.py
@@ -1,6 +1,8 @@
 from __future__ import with_statement
 
 import atexit
+import tempfile
+import os
 import unittest
 from datetime import datetime
 import flask
@@ -384,12 +386,10 @@ class PaginationTestCase(unittest.TestCase):
 class BindsTestCase(unittest.TestCase):
 
     def test_basic_binds(self):
-        import tempfile
         _, db1 = tempfile.mkstemp()
         _, db2 = tempfile.mkstemp()
 
         def _remove_files():
-            import os
             try:
                 os.remove(db1)
                 os.remove(db2)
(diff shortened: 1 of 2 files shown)


test_sqlalchemy.py
```diff
@@ -456,6 +456,44 @@ class Baz(db.Model):
...
+        def _remove_files():
+            try:
+                os.remove(db1)
+                os.remove(db2)
+            except IOError:
+                pass
+        atexit.register(_remove_files)
+
+        app = flask.Flask(__name__)
+        app.config['SQLALCHEMY_ENGINE'] = 'sqlite://'
```
davidism: should be `SQLALCHEMY_DATABASE_URI`

test_sqlalchemy.py
```diff
@@ -456,6 +456,44 @@ class Baz(db.Model):
...
+        _, db1 = tempfile.mkstemp()
+        _, db2 = tempfile.mkstemp()
+
+        def _remove_files():
+            try:
+                os.remove(db1)
+                os.remove(db2)
+            except IOError:
+                pass
+        atexit.register(_remove_files)
```
davidism: There's no need for creating and removing db files, just use `sqlite://` for all binds, each will connect to a different in-memory db.

test_sqlalchemy.py
```diff
@@ -456,6 +456,44 @@ class Baz(db.Model):
...
+
+        app = flask.Flask(__name__)
+        app.config['SQLALCHEMY_ENGINE'] = 'sqlite://'
+        app.config['SQLALCHEMY_BINDS'] = {
+            'foo':      'sqlite:///' + db1,
+            'bar':      'sqlite:///' + db2
```
davidism: This bind is never used.

test_sqlalchemy.py
```diff
@@ -456,6 +456,44 @@ class Baz(db.Model):
...
+
+        class FooBoundModel(AbstractFooBoundModel):
+            id = db.Column(db.Integer, primary_key=True)
+
+        db.create_all()
+
+        # does the model have the correct engines?
+        self.assertEqual(db.metadata.tables['foo_bound_model'].info['bind_key'], 'foo')
+
+        # see the tables created in an engine
```
davidism: This section seems unnecessary.  Could be replaced with `db.get_engine(bind='foo').execute('select * from users')`, which would raise if the table didn't exist, but we're not really testing that binds work here.

davidism:
Thanks.  Please address the review comments and this can be merged.

davidism:
OK, realized that you were just copying the other test cases, so I'll go clean that up myself.

singingwolfboy:
Ah, thank you. I meant to point that out in the pull request, and then never got around to it.

davidism:
cleaned up in #388

PR #376, opened 2016-02-23
Pessimistic disconnection handling. Resolves #375

### LINKED ISSUE #375: Pessimistic disconnection handling
Whenever a database connection dies while it is in the connection pool (ie. because the database server is restarted), the next query made will throw an exception and most-likely end up in a HTTP 500 for this request. This seems to be by design in SQLAlchemy it's what they call "optimistic disconnection handling". (see [Disconnection handling](http://docs.sqlalchemy.org/en/rel_1_0/core/pooling.html?highlight=reconnect#dealing-with-disconnects) in the documentation).

In the documentation they also provide a way to check the connection before it is used. This has a tiny performance implications but avoids most such errors.

- What is the best way to use "pessimistic disconnection handling" with Flask-SQLAlchemy?
- Should this be built into Flask-SQLAlchemy?
- Should it be the default behavior?

torotil:
See issue #375.

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -22,7 +22,7 @@ from flask import _request_ctx_stack, abort, has_request_context, request
 from flask.signals import Namespace
 from operator import itemgetter
 from threading import Lock
-from sqlalchemy import orm, event, inspect
+from sqlalchemy import orm, event, inspect, select
 from sqlalchemy.orm.exc import UnmappedClassError
 from sqlalchemy.orm.session import Session as SessionBase
 from sqlalchemy.engine.url import make_url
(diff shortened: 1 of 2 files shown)


davidism:
This needs tests and docs before merging.  I'm also not sure if this needs to be merged/supported here.  If someone needs this, it's trivial to implement, they just copy paste the recipe from the docs.

torotil:
@davidism:

1. It might be trivial to you. But it is far from trivial for me.
2. This is an issue that comes up again and again on SO and on the sqlalchemy-mailinglist. It's sad that there is no out-of-the-box solution.
3. Just adding this to every project increases the amount of boilerplate needed for using sqlalchemy with flask.
4. Testing is difficult: You need to somehow make the connection fail without sqlalchemy noticing it.

torotil:
I'm closing this for now until there is a decision how to approach this. This PR was just a way to show goodwill anyway.

PR #387, opened 2016-04-09
[Update] support slave read

fatelei:
+ add `with_slave` decorator to support master-slave model.


davidism:
Using globals like this is not the way to go.  Also this includes no tests or documentation.

PR #388, opened 2016-04-09
Clean up tests, more thorough version testing

davidism:
* `SQLALCHEMY_ENGINE` isn't a config key, so the default `sqlite://` was being used.
* `SQLALCHEMY_DATABASE_URI` defaults to `sqlite://` anyway, so just remove all these configs
* `CustomQueryClassTestCase` wasn't registered
* `CustomQueryClassTestCase.test_dont_override_model_default` wasn't testing the right thing, and was also failing
* `CustomModelClassTestCase` wasn't registered
* `import flask_sqlalchemy as sqlalchemy` is a confusing alias, used `fsa` instead
* added "lowest" and "current" tox envs for testing dependency versions
* Flask lowest tests against blinker==1.0, which doesn't have `connected_to`, replace with `connect` and `disconnect`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,11 +9,15 @@ python:
   - "3.4"
   - "3.5"
 
+env:
+  - REQUIREMENTS=lowest
+  - REQUIREMENTS=current
+
 install:
   - pip install tox
 
 script:
-  - tox -e py
+  - tox -e py-$REQUIREMENTS
 
 notifications:
   email: false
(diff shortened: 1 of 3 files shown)


davidism:
Ignore the failed ci build, I think it got confused between the initial push and the pr.  It does pass: https://travis-ci.org/mitsuhiko/flask-sqlalchemy/builds/121980819.

PR #393, opened 2016-05-02
Make paginate more efficient with many pages.

dourvaris:
Current paginate iterates over every single page possible, and discards most of them, for large page sets > 50,000 this can be quite slow and noticeable.

This patch should make paginating on many pages much faster since it won't iterate over every single page possible, it should make it constant time more or less.

Speed for 100,000 pages:

``` ipython
(old)
In [14]: %timeit list(flask_sqlalchemy.Pagination(None, 4, 10, 1000000, []).iter_pages())
10 loops, best of 3: 81.6 ms per loop

(new)
In [11]: %timeit list(flask_sqlalchemy.Pagination(None, 4, 10, 1000000, []).iter_pages())
100000 loops, best of 3: 11 µs per loop

```

Also added a bunch of tests for edge cases.

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -390,8 +390,20 @@ class Pagination(object):
               </div>
             {% endmacro %}
         """
+
+        pages = []
+        pages += range(1, min(left_edge + 1, self.pages + 1))
+        pages += range(
+            max(self.page - left_current, left_edge + 1),
+            min(self.page + right_current, self.pages - right_edge)
+        )
+        pages += range(
+            max(self.pages - right_edge, right_edge + 1),
+            self.pages + 1
+        )
+
         last = 0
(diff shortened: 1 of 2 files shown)


flask_sqlalchemy/__init__.py
```diff
@@ -390,16 +390,28 @@ def iter_pages(self, left_edge=2, left_current=2,
               </div>
             {% endmacro %}
         """
+
+        pages = []
+        pages += range(
```
justanr: This breaks for Python 3. Why not `pages.extend` or  just `itertools.chain(...)` since there's not a reason to keep them all in memory anyways.

dourvaris:
Is there any interest in this pr at all?

dhhagan:
Has/is this going to be merged..?

davidism:
fixed in #1087

Issue #401, opened 2016-06-09
MariaDB defaults cause `2013, 'Lost connection to MySQL server during query'`


nickw444:
Whilst not directly an issue with this library, it might be worth noting that the default that is chosen for `SQLALCHEMY_POOL_RECYCLE` when using MariaDB as a drop in replacement for MySQL causes a hard to debug connection error.

Comparing the defaults between MYSQL and MariaDB we find:
## MySQL `5.7.12`:

```
mysql> SHOW GLOBAL VARIABLES LIKE "wait_timeout";
+---------------+-------+
| Variable_name | Value |
+---------------+-------+
| wait_timeout  | 28800 |
+---------------+-------+
1 row in set (0.00 sec)
```
## MariaDB `10.1.14-MariaDB-1~xenial`

```
MariaDB [(none)]> SHOW GLOBAL VARIABLES LIKE "wait_timeout";
+---------------+-------+
| Variable_name | Value |
+---------------+-------+
| wait_timeout  | 600   |
+---------------+-------+
1 row in set (0.00 sec)

root@c554e4589da8:/# cat /etc/mysql/my.cnf | grep timeout
connect_timeout     = 5
wait_timeout        = 600
```

Flask-SQLAlchemy defaults to using a 2 hour recycle time (`7200 seconds`), which obviously is too long for MariaDB's default `600 second` timeout, however is a reasonable value for MySQL, which defaults to `28800 seconds`. 

Personally, I think it might be worth noting this in the [configuration keys](http://flask-sqlalchemy.pocoo.org/2.1/config/#configuration-keys) page of the documentation for the `SQLALCHEMY_POOL_RECYCLE` key, since that is the first place users are generally directed when searching the issue.

davidism:
Is there a way to detect Maria vs MySQL during `apply_driver_hacks`?

nickw444:
Unless the user provides a mariadb specific database URI that we re-write to MySQL, I can't think of a way. Plus this also takes away from mariadb being a drop in replacement. I suppose we could run the select query above to calculate the timeout on initialisation? 

But I think that's putting too much responsibility into this library. It really should come down to user configuration.

immunda:
Whilst a valid suggestion @nickw444, I think you're right that it's better to avoid confusing the issue with a `mariadb` URI.

[...]

PR #416, opened 2016-08-13
Provide additional documentation for SQLALCHEMY_POOL_RECYCLE

nickw444:
From issue #401, adding additional documentation to help users configuring Flask-SQLAlchemy with a backends with a shorter idle connection timeout such as:
- MariaDB
- SAAS/IAAS provided databases that are pre-configured

I've included the error that generally indicates this is an issue `2013: Lost connection to MySQL server during query` so users googling the exception will have a better chance of finding appropriate documentation.


immunda:
Thanks sir!

PR #432, opened 2016-09-19
SQLALCHEMY_POOL_SIZE cannot be used with NullPool strategy -> removin…

jboegeholz:
…g parameter from config

SQLAlchemy defaults to a NullPool Strategy and throws an error when pool_size is set.
This fix pops the param from the config and issues a warning
(Principle of least surprise)


rsyring:
Currently, FSA will use the NullPool only if pool_size is None or set to 0.  I don't think this PR is relevant anymore.  If you still believe there is a problem, please open an issue to discuss.  Thanks.

PR #437, opened 2016-10-04
Use same tense uniformly in all docstrings

akki:
Just noticed this while going through the code.


davidism:
This is mode, not tense.  I agree, function docs should be indicative, not imperative.  (Instructions / tutorials would be imperative.)  Thanks for bringing attention to it, it's the first time I've thought about the distinction in docs.

akki:
Your welcome. It's always great to be consistent so I thought of sending
this PR.

Akshesh (akki)
IIT Roorkee

davidism:
Ended up deciding against this. Thanks anyway.

PR #441, opened 2016-10-18
Have sqlite relative paths use the config's root

vault:
This makes relative paths do the right thing when the application has
instance_relative_config set to True.

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -887,7 +887,7 @@ class SQLAlchemy(object):
 
             # if it's not an in memory database we make the path absolute.
             if not detected_in_memory:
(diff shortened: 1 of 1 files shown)


davidism:
This will fail if the instance folder doesn't exist.

vault:
I'm not sure I understand that concern. Wouldn't that be true for most things when `instance_relative_config` is True? 

e.g. I have an app with that set, and call `app.config.from_pyfile(relative_path)` it will also fail if the instance folder doesn't exist.

davidism:
You can set `silent=True` and it won't.  Presumably you'd have a default config and the local config is optionally in the instance folder.

vault:
Ah, I see the problem now. Got blinded a bit by my use-case which is to consume a read-only database that gets dropped in the instance path.

Issue #421, opened 2016-08-28
issue warning if SQLALCHEMY_DATABASE_URI is not set


scotchka:
Currently, if the user misspells SQLALCHEMY_DATABASE_URI, say, as SQLALCHEMY_DATABASE_URL, the database defaults to SQLite which is usually not the intended database. It would be better to raise a UserWarning that makes clear the the app is defaulting to SQLite.

Any reason to not issue such a warning? If not, I can make the pull request.

kingsj0405:
I really agree with this issue... I spent over 5 hours realizing URI not URL...

PR #443, opened 2016-10-29
warn user when SQLALCHEMY_DATABASE_URI is not set

scotchka:
https://github.com/mitsuhiko/flask-sqlalchemy/issues/421


davidism:
I updated the check to reflect what `setdefault` is doing and to explicitly use `sqlite:///:memory:` for clarity. Also rebased against master.

PR #446, opened 2016-11-15
Update __init__.py

kecheon:
calculating method for Pagination.total changed.
When it comes to query.count() is greater than len(query.all()),
pagination does not work correctly.


davidism:
This issues a second query, as opposed to getting the length of the already fetched page of items.

PR #447, opened 2016-11-21
Add `db.transaction` block

### LINKED ISSUE #217: Add `db.transaction` block
Safely commits if no errors, will rollback otherwise.

fantix:
This fixed #217. Please let me know if the commits need to be squashed.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -63,6 +63,19 @@ A list of configuration keys currently understood by the extension:
                                    that it will be disabled by default in
                                    the future.  This requires extra memory
                                    and should be disabled if not needed.
+``SQLALCHEMY_ISOLATE_TRANSACTION`` Controls the default behavior of root
+                                   transactions whether to isolate boundary
+                                   to definition. Default is ``True``, means
+                                   entering `with db.transaction()` with a
+                                   session with `autocommit=False` will cause
+                                   a rollback first. Oppositely ``False`` will
+                                   stop the entering rollback from happening.
+``SQLALCHEMY_NESTED_TRANSACTION``  Controls the default implementation of
+                                   nested transactions, ``False`` to use
+                                   programed subtransactions (default, the
+                                   same as ``begin(subtransactions=True)`` in
+                                   SQLAlchemy), ``True`` to use DB savepoint
+                                   (the same as ``begin_nested``).
 ================================== =========================================
 
 .. versionadded:: 0.8
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -28,6 +28,7 @@ from sqlalchemy.orm.exc import UnmappedClassError
 from sqlalchemy.orm.session import Session as SessionBase
 from sqlalchemy.engine.url import make_url
 from sqlalchemy.ext.declarative import declarative_base, DeclarativeMeta
+from werkzeug.local import LocalStack
 from flask_sqlalchemy._compat import iteritems, itervalues, xrange, string_types
 
 # the best timer function for the platform
@@ -737,6 +738,8 @@ class SQLAlchemy(object):
         self.use_native_unicode = use_native_unicode
         self.Query = query_class
         self.session = self.create_scoped_session(session_options)
+        self.stack = LocalStack()
+        self.stack.__ident_func__ = lambda: id(self.session())
         self.Model = self.make_declarative_base(model_class, metadata)
         self._engine_lock = Lock()
         self.app = app
@@ -917,14 +920,13 @@ class SQLAlchemy(object):
     @contextmanager
     def transaction(self, isolate=None, nested=None, **kwargs):
         session = self.session()
-        stack = session.info.setdefault('sa_stack', deque())
-        is_root = len(stack) == 0
+        is_root = self.stack.top is None
 
         if is_root:
             nested = False
             item = {}
         else:
-            item = stack[-1].copy()
+            item = self.stack.top.copy()
 
         if nested is None:
             nested = self.get_app().config['SQLALCHEMY_USE_NESTED_TRANSACTION']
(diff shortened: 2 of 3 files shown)


davidism:
This seems related to #358 "Add support for nested sessions (SAVEPOINTs) in modifications tracking", which I'd like to get merged too. If you're interested, could you take a look and see how this would affect that?

fantix:
Oh sure, thanks for the quick response! I'll check on that and maybe get this PR an update.

fantix:
Reviewed #358, should be fine for both to work together. Still glad you mentioned #358 because two updates can be done here:

1. The same SQLite workaround in test is applied slightly differently, I shall apply the style of #358 here;
2. Setting private attribute on session directly is a current practice, I shall follow the same way here, getting rid of `LocalStack` (a previous commit used `session.info` which is not present in minimum requirement).

Will update PR soon.

roganov:
Could you please explain when to use this over `db.session.begin()`?

fantix:
@roganov Generally speaking, `with db.transaction()` is preferred in most cases where db access happens, because *explicit is better than implicit* (harder to leave a transaction unclosed locking resources for a long time), and `db.transaction()` also takes care of a bit more magic underneath unifying the way transactions are defined. In all three scenarios:

With the default `autocommit=False`, directly calling `db.session.begin()` will fail saying `A transaction is already begun.  Use subtransactions=True to allow subtransactions.`. According to SQLAlchemy [document](http://docs.sqlalchemy.org/en/rel_1_1/orm/session_transaction.html) the session is by default in a begin state, convenient for emitting queries directly. But when it comes to transactions, it is a good practice to explicitly define where the transaction truly begins, so that previous unexpected queries won't be included accidentally in the transaction, ruining consistency for example. So to guarantee a clean isolated transaction, `db.session.rollback()` is instead the right thing to do here - which is the default behavior of `with db.transaction()`. (Of course you can choose not to do so, in both ways.)

Secondly, one could do `db.session.begin(subtransactions=True)`, followed by two `db.session.commit()` - one to close the subtransaction (no `COMMIT` in db), another to actually commit the db transaction. This is confusing and gets messy easily. It is in this case preferred to do something like this:

```python
with db.transaction():               # ROLLBACK
    db.session.execute('SQL 1')      # BEGIN (implicit); SQL 1;
    with db.transaction():           # no SQL here, push one subtransaction to stack
        db.session.execute('SQL 2')  # SQL 2
                                     # no SQL here, pop subtransaction from stack
    db.session.execute('SQL 3')      # SQL 3
                                     # COMMIT or ROLLBACK;
```

The same logic without `db.transaction`:

```python
db.session.rollback()                       # ROLLBACK;
try:
    db.session.execute('SQL 1')             # BEGIN (implicit); SQL 1;
    db.session.begin(subtransactions=True)  # push
    try:
        db.session.execute('SQL 2')         # SQL 2;
        db.session.commit()                 # pop
    except:
        db.session.rollback()               # pop and mark failed
        raise
    db.session.execute('SQL 3')             # SQL 3;
    db.session.commit()                     # COMMIT;
except:
    db.session.rollback()                   # ROLLBACK;
    raise
```

Finally, if `autocommit` is set to `True` (which is not a recommended setting), it is preferred to use all the same syntax with `db.transaction()` to the same logic - in other words, the same code will support both `autocommit=False` and `autocommit=True` (please notice the difference of SQLs than the first example):

```
# db.session.autocommit = True
with db.transaction():               # BEGIN;
    db.session.execute('SQL 1')      # SQL 1;
    with db.transaction():           # no SQL here, push one subtransaction to stack
        db.session.execute('SQL 2')  # SQL 2
                                     # no SQL here, pop subtransaction from stack
    db.session.execute('SQL 3')      # SQL 3
                                     # COMMIT or ROLLBACK;
```

lelabo-marc:
Any new on this ? I find this method much clearer than using db.session.begin() and rollbacks.

ticosax:
this pull request is outstanding.
It provides a clever decoupling between the business logic and administrative task of dealing with databases transactions within the views. It is more reliable and closer to the semantics of dealing with transactional databases by tying the http request to the database transaction. HTTP success -> commit(), HTTP failure -> rollback(). For data critical businesses, this is a must to have.

As an inspiration, Django took the same approach https://docs.djangoproject.com/en/2.1/topics/db/transactions/#tying-transactions-to-http-requests

rsyring:
Adding this to the 3.0 milestone.  Looks like it's going to need some rebasing for at least the test suite.

ticosax:
@rsyring when is 3.0 expected ? what is the roadmap ?
I volunteer to take care of this pull request unless @fantix still wants to do it.

fantix:
@ticosax please feel free to take over, I appreciate that 😃

rsyring:
@ticosax our milestones are up to date and we also have an issue discussing release plans:

https://github.com/pallets/flask-sqlalchemy/milestones
https://github.com/pallets/flask-sqlalchemy/issues/682

But there are no date commitments.  I'm just starting to contribute to the project, so some of it is going to depend on how much time I have to give (I'm shooting for a few days a month) and how fast PRs and releases get turned around (which I'm waiting on other maintainers for).

ticosax:
@fantix thanks for your quick feedback, then I'll do my best to upgrade it.
@rsyring Thanks for the link, I have a lot to catch up.

davidism:
There doesn't appear to be anything to do here besides a rebase onto master, which maintainers can do right before merging. Is there something specific you think still needs work @ticosax.

ticosax:
@davidism no I don't plan to do more. than having passing tests on top of current master.
I can't tell if the branch can be rebased automatically since I don't write permission. So I'm offering my help to make sure it is mergeable. If it can be merged as it is, that's even better.
Please, Let me know if I should do it or not.

davidism:
Looks like the tests need to be rewritten to use the pytest style and fixtures rather than unittest.

ticosax:
alright, then I'll make sure they can run with pytest. thanks

davidism:
Continued in #695

PR #459, opened 2017-01-13
Stop recommending lazy='dynamic' and other BadThings™ new users tend to keep

ThiefMaster:


docs/models.rst
```diff
@@ -76,14 +72,14 @@ function.  However the foreign key has to be separately declared with the
 
     class Person(db.Model):
         id = db.Column(db.Integer, primary_key=True)
-        name = db.Column(db.String(50))
-        addresses = db.relationship('Address', backref='person',
-                                    lazy='dynamic')
+        name = db.Column(db.String(50), nullable=False)
+        addresses = db.relationship('Address', backref='person', lazy=True)
```
davidism: Don't need `lazy=True`, that's the default. (Well, `'select'` is the default, `True` is a synonym.)

docs/quickstart.rst
```diff
@@ -68,6 +64,13 @@ Accessing the data in database is easy as a pie::
     >>> User.query.filter_by(username='admin').first()
     <User u'admin'>
 
+Note how we never defined a ``__init__`` method on the `User` class?
+That's because SQLAlchemy adds an implicit constructor to all model
+classes which accepts keyword arguments for all its columns and
+relationships.  If you decide to override the constructor for any
+reason, make sure to keep acceptign ``**kwargs`` and call the super
```
dsm054: typo: accepting

docs/models.rst
```diff
@@ -76,14 +72,14 @@ function.  However the foreign key has to be separately declared with the
...
 
     class Address(db.Model):
         id = db.Column(db.Integer, primary_key=True)
-        email = db.Column(db.String(50))
-        person_id = db.Column(db.Integer, db.ForeignKey('person.id'))
+        email = db.Column(db.String(120), nullable=False)
+        person_id = db.Column(db.Integer, db.ForeignKey('person.id'),
```
davidism: Don't need to specify the column type for foreign keys, it's inferred from the linked column.

docs/models.rst
```diff
@@ -127,19 +135,29 @@ helper table that is used for the relationship.  For this helper table it
 is strongly recommended to *not* use a model but an actual table::
 
     tags = db.Table('tags',
-        db.Column('tag_id', db.Integer, db.ForeignKey('tag.id')),
-        db.Column('page_id', db.Integer, db.ForeignKey('page.id'))
+        db.Column('tag_id', db.Integer, db.ForeignKey('tag.id'), nullable=False),
+        db.Column('page_id', db.Integer, db.ForeignKey('page.id'), nullable=False)
```
davidism: Usually I mark both columns as `primary_key=True`. Same comment about not needing type applies as well.

docs/models.rst
```diff
@@ -93,31 +89,43 @@ default from your declaration.  If you would want to have a one-to-one
...
+-   ``'joined'`` / ``False`` tells SQLAlchemy to load the relationship in
+    the same query as the parent using a `JOIN` statement.
 -   ``'subquery'`` works like ``'joined'`` but instead SQLAlchemy will
     use a subquery.
--   ``'dynamic'`` is special and useful if you have many items.  Instead of
-    loading the items SQLAlchemy will return another query object which
-    you can further refine before loading the items.  This is usually
-    what you want if you expect more than a handful of items for this
-    relationship.
+-   ``'dynamic'`` is special and can be useful if you have many items.
```
davidism: This might be expanded on to say "if you have many items and typically need a subset of them" or something similar.

docs/models.rst
```diff
@@ -127,19 +135,29 @@ helper table that is used for the relationship.  For this helper table it
...
+            backref=db.backref('pages', lazy=True))
 
     class Tag(db.Model):
         id = db.Column(db.Integer, primary_key=True)
 
-Here we configured `Page.tags` to be a list of tags once loaded because we
-don't expect too many tags per page.  The list of pages per tag
-(`Tag.pages`) however is a dynamic backref.  As mentioned above this means
-that you will get a query object back you can use to fire a select yourself.
+Here we configured `Page.tags` to be loaded immediately after loading
```
davidism: Use double backticks for code in rst.

docs/models.rst
```diff
@@ -127,19 +135,29 @@ helper table that is used for the relationship.  For this helper table it
...
+a Page, but using a separate query.  This always results in two
+queries when retrieving a Page, but when querying for multiple pages
+you will not get additional queries.
+
+The list of pages for a tag on the other hand is something that's
+rarely needed - for sure not when you just get the tags e.g. for a
```
davidism: "for sure" and "e.g." sounds awkward, perhaps break this up into two sentences. "rarely needed. For example, you don't need them when getting the tags for a specific page."

docs/models.rst
```diff
@@ -127,19 +135,29 @@ helper table that is used for the relationship.  For this helper table it
...
+specific page.  For this reason the backref is set to be lazy-loaded.
+That way accessing it for the first time will trigger a query to get
+the list of pages for that tag.  If you need to apply further query
+options on that list, you could either switch to the ``'dynamic'``
+strategy - with the drawbacks mentioned above - or get a query object
+using ``Page.query.with_parent(some_tag)`` and then use it exactly
```
davidism: Should we show a concrete example of this type of query?

docs/models.rst
```diff
@@ -127,19 +135,29 @@ helper table that is used for the relationship.  For this helper table it
...
-that you will get a query object back you can use to fire a select yourself.
+Here we configured `Page.tags` to be loaded immediately after loading
+a Page, but using a separate query.  This always results in two
+queries when retrieving a Page, but when querying for multiple pages
+you will not get additional queries.
+
+The list of pages for a tag on the other hand is something that's
+rarely needed - for sure not when you just get the tags e.g. for a
+specific page.  For this reason the backref is set to be lazy-loaded.
+That way accessing it for the first time will trigger a query to get
```
davidism: Combine with previous sentence. "Therefore, the backref is set to be lazy-loaded so that accessing it for the first time..."

docs/quickstart.rst
```diff
@@ -68,6 +64,13 @@ Accessing the data in database is easy as a pie::
     >>> User.query.filter_by(username='admin').first()
     <User u'admin'>
 
+Note how we never defined a ``__init__`` method on the `User` class?
+That's because SQLAlchemy adds an implicit constructor to all model
+classes which accepts keyword arguments for all its columns and
+relationships.  If you decide to override the constructor for any
+reason, make sure to keep accepting ``**kwargs`` and call the super
+constructor with those ``**kwargs`` to preserve this behavior::
```
davidism: Looks like there was supposed to be an example following this.

docs/quickstart.rst
```diff
@@ -80,53 +83,58 @@ application that uses two tables that have a relationship to each other::
...
+not matter whether ``db.session.add()`` is called before or after
+creating these objects.  The association can also be done on either
+side of the relationship - so a post can be created with a category
+or it can be added to the list of posts of the category.
+
+Let's look at the posts. Accessing them will load them from the database
+since the relationship is lazy-loaded, but you will probably not notice
+the difference - loading a list is quite fast. Just don't do it when
+looping over many objects (in such a case you would rather use the
+``'subquery'`` or ``'joined'`` loading strategy in the model or as a
```
davidism: Might be helpful to show an example of when and how to use the `joinedload` option.

docs/models.rst
ThiefMaster: I prefer to be explicit ;)

docs/models.rst
ThiefMaster: really? interesting. didn't know about that.. not sure what i prefer there though, but maybe i'm just used to specifying it

docs/models.rst
```diff
@@ -127,19 +135,29 @@ helper table that is used for the relationship.  For this helper table it
...
+the list of pages for that tag.  If you need to apply further query
+options on that list, you could either switch to the ``'dynamic'``
+strategy - with the drawbacks mentioned above - or get a query object
+using ``Page.query.with_parent(some_tag)`` and then use it exactly
+as you would with the query object from a dynamic relaationship.
```
davidism: typo: relationship has an extra a

docs/models.rst
davidism: http://docs.sqlalchemy.org/en/latest/core/metadata.html#sqlalchemy.schema.Column.params.type_ Apparently it's been in there since before 0.9, but I just discovered it recently too.

docs/models.rst
ThiefMaster: so do i, didn't even notice that there's no PK :o

docs/models.rst
davidism: I noticed the comment further down, I'm fine leaving it. Although some of the backrefs aren't explicit.

docs/models.rst
ThiefMaster: I included one somewhere else (close to the old example for dynamic queries)

docs/quickstart.rst
ThiefMaster: Yeah I was going to add one but then forgot it

docs/models.rst
ThiefMaster: Just checked, all `backref()` ones are explicit, for simple `backref='name'` cases there's no way to be more explicit without switching to `backref()`

docs/quickstart.rst
ThiefMaster: example added

docs/quickstart.rst
ThiefMaster: added

docs/models.rst
ThiefMaster: 0.8 is still supported :/ If it was unreliable there it's probably better to keep the type here for now and remove it once support for ancient SQLAlchemy Versions is dropped. FWIW, I think a Flask-SQLAlchemy==3.0 should switch to SQLAlchemy 1.0 at least, or even better 1.1.

davidism: Yes, we're going to depend on at least 1.0 for 3.0, so we can update this then.

davidism:
There are still incorrect single backticks in your changes and the rest of the file, would you mind giving the whole file a quick check?

ThiefMaster:
Hopefully now all of them are gone

davidism:
Two left as far as I saw:

- tablename on models line 22
- User on quickstart line 67

Issue #323, opened 2015-07-02
Adding @declared_attr __tablename__ to Mixin is overridden by _BoundDeclarativeMeta


mike-hart:
Using 2.0

I was using a mixin on my models to add a `__tablename__` method to work out a prefixed/namespaced table name based on the class name but `_BoundDeclarativeMeta`'s doesn't appear to check against base classes before adding its own.

This lead to errors like:
`NoReferencedTableError: Foreign key associated with column 'not_yet_built.dependent_thoroughfare_id' could not find table 'paf_thoroughfare' with which to generate a foreign key to target column 'thoroughfare_id'`

``` python
class PAFMixin(object):

    @declared_attr
    def __tablename__(cls):
        return 'paf' + re.sub(
            '[A-Z]', lambda match: '_' + match.group(0).lower(), cls.__name__)
```

Note that the class would prefix the table-name with paf, but this was not present in the first name of the exception (the latter is the reference from the `ForeignKey` itself). This prevents namespacing on systems that don't support schema (MySQL) and could be very confusing for others.

davidism:
Would you produce a full test case for this?  Also, ensure that you are using the latest version of SQLAlchemy 1.0, 0.9, or 0.8 branches, and specify which you're using.

mike-hart:
```python
import re

from flask import Flask
from flask.ext.sqlalchemy import SQLAlchemy
from sqlalchemy import (Column, ForeignKey, Integer)
from sqlalchemy.ext.declarative import declared_attr


app = Flask(__name__)
app.config['SQLALCHEMY_DATABASE_URI'] = 'sqlite://'
db = SQLAlchemy(app)


class PAFMixin(object):

    @declared_attr
    def __tablename__(cls):
        return 'paf' + re.sub(
            '[A-Z]', lambda match: '_' + match.group(0).lower(), cls.__name__)


class Locality(PAFMixin, db.Model):

    locality_id = Column(Integer, primary_key=True)


class Address(PAFMixin, db.Model):

    address_id = Column(Integer, primary_key=True)
    locality_id = Column(Integer, ForeignKey('paf_locality.locality_id'))


db.create_all()
````

Results in:
`sqlalchemy.exc.NoReferencedTableError: Foreign key associated with column 'address.locality_id' could not find table 'paf_locality' with which to generate a foreign key to target column 'locality_id'`

```
$ pip freeze | grep Alc
Flask-SQLAlchemy==2.0
[...]

PR #467, opened 2017-02-27
Don't access all class properties when generating tablename

### LINKED ISSUE #349: _should_set_tablename is problematic
It runs at import time and ends up accessing _all_ attributes on the model classes.

My base model class has a class-level property like this:

```
@classproperty
@classmethod
def has_rows(cls):
    """Checks if the underlying table has any rows.

    This is done in an efficient way and should preferred over
    calling the `count` method unless you actually care about
    the exact number of rows.
    """
    from indico.core.db import db
    # we just need one "normal" column so sqlalchemy doesn't involve relationships
    # it doesn't really matter which one it is - it's never even used in the query
    pk_col = getattr(cls, inspect(cls).primary_key[0].name)
    return db.session.query(db.session.query(pk_col).exists()).one()[0]
```

Updating to 2.1 breaks since the local import inside that method ends up being executed as import time again; also, accessing it would trigger SQL if it didn't fail.

I think in this case I'll simply convert it to a normal classmethod, but I wonder if there are any other cases that might break...

davidism:
When examining model to generate tablename, only get `declared_attr` properties against the model, not all class properties.

Move tablename generation to the base `Model` to avoid SQLAlchemy warning about unmanaged access to declared attrs. (One test still raises this because abstract classes are unmanaged but still generate a name.)

`_BoundDeclarativeMeta` only handles popping off tablenames if they are manually set, so that subclasses still autogenerate. One side effect of this is that custom base models must inherit from the default base model (or re-implement the `__tablename__` declared attr), otherwise it would be impossible to set tablenames.

Thanks to @vToMy for helping come up with fixes. This patch is mostly based off their patch in https://github.com/mitsuhiko/flask-sqlalchemy/issues/349#issuecomment-272960086.

closes #349, #323, #454

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -25,11 +25,12 @@ from flask import _app_ctx_stack, abort, current_app, g, request
 from flask.signals import Namespace
 from sqlalchemy import event, inspect, orm
 from sqlalchemy.engine.url import make_url
-from sqlalchemy.ext.declarative import DeclarativeMeta, declarative_base
+from sqlalchemy.ext.declarative import DeclarativeMeta, declarative_base, \
+    declared_attr
 from sqlalchemy.orm.exc import UnmappedClassError
 from sqlalchemy.orm.session import Session as SessionBase
 
-from ._compat import itervalues, string_types, xrange
+from ._compat import iteritems, itervalues, string_types, xrange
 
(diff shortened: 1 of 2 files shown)


davidism:
Not sure how to add a test for "doesn't access class properties", since dict iteration is unordered. If anyone has a good test for this, another patch would be appreciated.

PR #472, opened 2017-02-27
Suppress warning about not set DATABASE_URI when SQLALCHEMY_BINDS are set

### LINKED ISSUE #471: Do not warn about SQLALCHEMY_DATABASE_URI when SQLALCHEMY_BINDS is set
Currently running following test case:
```python
import flask
import flask_sqlalchemy as fsa

app = flask.Flask(__name__)
    'foo': 'sqlite://',
    'bar': 'sqlite://'
}
db = fsa.SQLAlchemy(app)

class Foo(db.Model):
    __bind_key__ = 'foo'
    __table_args__ = {"info": {"bind_key": "foo"}}

class Bar(db.Model):
    __bind_key__ = 'bar'

class Baz(db.Model):

db.create_all()
```
Raises:
```
/usr/lib64/python3.5/site-packages/flask_sqlalchemy/__init__.py:819: UserWarning: SQLALCHEMY_DATABASE_URI not set. Defaulting to "sqlite:///:memory:".
  'SQLALCHEMY_DATABASE_URI not set. Defaulting to '
```

I think that it is a good idea to let user know that he or she probably forgotten/misspelled database URI, but in my opinion setting SQLALCHEMY_BINDS should surpass such warning.

krassowski:
Fixes #471 at https://github.com/mitsuhiko/flask-sqlalchemy/issues/471.

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -814,10 +814,13 @@ class SQLAlchemy(object):
         of an application not initialized that way or connections will
         leak.
         """
-        if 'SQLALCHEMY_DATABASE_URI' not in app.config:
+        if (
+            'SQLALCHEMY_DATABASE_URI' not in app.config and
+            'SQLALCHEMY_BINDS' not in app.config
+        ):
             warnings.warn(
-                'SQLALCHEMY_DATABASE_URI not set. Defaulting to '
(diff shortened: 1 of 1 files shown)


davidism:
On the other hand, this is somewhat misleading now, as `SQLALCHEMY_DATABASE_URI` still gets a default.

Issue #482, opened 2017-03-04
Make it easy to parse a Pagination object to dict


italomaia:
Being able to easely parse a Pagination object to dict would make it much more "ajax" friendly.

kalombos:
Hello, i want to contribute but i didn't understand what you need. You want to realize additional method at Pagination class  something like this:
```
def as_dct(self):
    return dict(items=self.items, page=self.page...)
```
?
Please give me more information and i will try to realize it. Thank you.

ThiefMaster:
Not `dct` for sure. That abbreviation is only OK when the name would be `dict` otherwise (and thus shadow the builtin). I'd go for something like `to_dict` or `as_dict`.

protream:
We can do it like this:

```
Class Pagination:
    ...

    def __getitem__(self, key):
        return getattr(self, key)

    def keys(self):
        return ('page', 'pages', 'total', 'has_prev', 'next_num', 'has_next')
```
Now, `dict` a Pagination Object jus do:

data = dict(pagination)
data['items'] = ...

spyoungtech:
Wouldn't it be most appropriate to implement a `__iter__` for this? That way one can simply do `dict(my_pagination_object)`

nicolashahn:
@protream's implementation as a PR: https://github.com/mitsuhiko/flask-sqlalchemy/pull/580

rtilk89:
@jab @davidism - I'm interested in giving this issue a shot, but looks like there is an unmerged PR. Is this still open for contributions?

tomowind:
@rtilk89 I was reading two unmerged PRs #485 and #580. #485 doesn't address comments fully, and #580 doesn't seem to incorporate comments in #485. So, you may extend #580 with tests and addressing comments in #485?

You can work on this one. I will do another.

davidism:
The problem with an `as_dict` method is that it does not guarantee that the result will be serializable by `jsonfiy` or `json.dumps`. Result rows can contain types that are not handled by the default JSON serializer, and it gets even worse when querying models, and it gets even worse when those models are not simple but use the multitude of features available to SQLAlchemy.

The direct way for a dev to handle this, although not the most powerful, is to implement their own `Flask.json_encoder` class to handle the things that would appear in `Pagination.items`. So if we do add a `to_dict` method, it needs to clearly explain that the developer still needs to do something else.

[...]

PR #485, opened 2017-03-20
added as_dict method for Pagination class

kalombos:
#482

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -404,9 +404,15 @@ class Pagination(object):
                 last = num
 
     def as_dict(self):
-        """Returns dict representation of the object"""
-        return dict(items=self.items, page=self.page, total=self.total,
-                    prev_num=self.prev_num, next_num=self.next_num)
+        """Create a dict with the pagination fields for use as a JSON response."""
+
+        return {
+            'items': self.items,
+            'page': self.page,
+            'total': self.total,
+            'prev_num': self.prev_num,
+            'next_num': self.next_num
+        }
 
 
 class BaseQuery(orm.Query):
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -403,6 +403,11 @@ class Pagination(object):
                 yield num
                 last = num
 
+    def as_dict(self):
+        """Returns dict representation of the object"""
+        return dict(items=self.items, page=self.page, total=self.total,
+                    prev_num=self.prev_num, next_num=self.next_num)
+
 
 class BaseQuery(orm.Query):
     """SQLAlchemy :class:`~sqlalchemy.orm.query.Query` subclass with convenience methods for querying in a web application.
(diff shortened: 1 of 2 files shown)


flask_sqlalchemy/__init__.py
```diff
@@ -403,6 +403,11 @@ def iter_pages(self, left_edge=2, left_current=2,
                 yield num
                 last = num
 
+    def as_dict(self):
+        """Returns dict representation of the object"""
+        return dict(items=self.items, page=self.page, total=self.total,
+                    prev_num=self.prev_num, next_num=self.next_num)
```
davidism: Use a dict literal here, not the callable, and use single-indent continuation style:

~~~python
return {
    'items': items,
    ...
}
~~~

flask_sqlalchemy/__init__.py
```diff
@@ -403,6 +403,11 @@ def iter_pages(self, left_edge=2, left_current=2,
                 yield num
                 last = num
 
+    def as_dict(self):
+        """Returns dict representation of the object"""
```
davidism: This needs a better doc string. Something like "Create a dict with the pagination fields for use as a JSON response."

test_sqlalchemy.py
```diff
@@ -385,6 +385,17 @@ def test_basic_pagination(self):
         self.assertEqual(list(p.iter_pages()),
                          [1, 2, None, 8, 9, 10, 11, 12, 13, 14, None, 24, 25])
 
+    def test_as_dict_method(self):
+        p = fsa.Pagination(None, 1, 20, 500, [1, 2, 3, 4, 5])
+        self.assertEqual(
+            p.as_dict(),
+            {'items': p.items,
+             'next_num': p.next_num,
```
davidism: Use single-indent continuation style, as demonstrated above.

davidism:
I'm worried about returning `items` in the dict. Since the use of this method is for generating a JSON response, this implies that the items will be JSON serializable, a guarantee that we do not make.

kalombos:
I've fixed code style but i don't know what to do with items. Just remove it from dictionary?

italomaia:
@kalombos if items always have to be properly parsed (meaning set again in the dictionary), then it probably makes sense to not have it in the output of to_dict. Example:

```
page = query.paginate()
result = page.to_dict()
result['items'] = schema.dump(page)
```

This syntax would also be quite nice:

```
page = query.paginate()
result = page.to_dict(schema.dump)
```

flask_sqlalchemy/__init__.py
italomaia: This was fixed.

flask_sqlalchemy/__init__.py
italomaia: This was fixed.

test_sqlalchemy.py
italomaia: This was fixed.

flask_sqlalchemy/__init__.py
```diff
@@ -403,6 +403,17 @@ def iter_pages(self, left_edge=2, left_current=2,
                 yield num
                 last = num
 
+    def as_dict(self):
+        """Create a dict with the pagination fields for use as a JSON response."""
```
davidism: Need to mention that the result of this will probably not be directly JSON serializable. To serialize query results and models, the dev will need to implement a custom `Flask.json_encoder` class.

davidism:
Hi, thanks for working on this! Based on further discussion in #482, I no longer think this is a feature we want to support right now, so I'm going to close it.

Issue #166, opened 2013-11-05
make it possible to pass additional options to create_engine


gromgull:
I needed to pass in an `isolation_level` parameters to the `create_engine` call. I did it like this: 

https://github.com/gromgull/flask-sqlalchemy/commit/689c90c48949dc4c020f16550576c517d2fb5230

(on top of the 1.0 tag since that's what we used, but I can make a pull-request with it applied to HEAD if you want)

jayzeng:
@immunda any update on this? we need to pass in a custom json json_serializer and need a way to specify/override engine_options, https://github.com/mitsuhiko/flask-sqlalchemy/pull/67/files looks good to me, any reason not to use that patch?

immunda:
This isn't going to be in the release 2.1 (very soon). However, it should definitely be addressed in v3.

@jayzeng It's not a problem with that commit per se, but there's a lot of details to sort out around documentation changes, tests and then managing the deprecation of options which such a solution would essentially override/render redundant.

jayzeng:
Completely understood. It is a pressing issue for us and I think we will just go for the subclass route. 

@immunda any rough estimate when v3 will be released?

jayzeng:
Also let me know what I can help to get it released asap

rknLA:
I'm also interested in some way to do this.  I need to [pass a custom connection function](http://docs.sqlalchemy.org/en/latest/core/pooling.html#using-a-custom-connection-function), and right now have to do some monkey patching around to get there.

Is there any value in working on or submitting a proposed fix at this point?

antoine-lizee:
+1 here too, in order to pass a custom `json_serializer`.

rsyring:
This really needs to be fixed.  It could be easily done by accepting a `SQLALCHEMY_ENGINE_OPTS` config value that is a dict and that would be merged into the options before being sent to create_engine().  That seems preferable to the implementation in #431 IMO.

This could literally be a two line change:

```python
            options = {'convert_unicode': True}
            self._sa.apply_pool_defaults(self._app, options)
            self._sa.apply_driver_hacks(self._app, info, options)
            if echo:
                options['echo'] = echo

            # next two lines are new
            config_engine_opts = self._app.config.get('SQLALCHEMY_ENGINE_OPTS', {})
            options.update(config_engine_opts)

            self._engine = rv = sqlalchemy.create_engine(info, **options)
```

[...]

PR #487, opened 2017-03-20
Allow engine options to be set through the constructor like session options.

keithjstone:
Package does not allow setting of custom engine options without monkey patching or inheriting the base class. This would allow options to be set and remembered prior to creating the engine in a blessed manner.

Our specific use-case is setting `use_native_uuids` to `False` on a postgres connection but I think this is more generally useful.

### DIFF
--- a/test_sqlalchemy.py
+++ b/test_sqlalchemy.py
@@ -503,6 +503,16 @@ class BindsTestCase(unittest.TestCase):
         self.assertEqual(len(metadata.tables), 1)
         self.assertTrue('foo_bound_model' in metadata.tables)
 
+    def test_extended_engine_options(self):
+        app = flask.Flask(__name__)
+        app.config['SQLALCHEMY_BINDS'] = {
+            'foo': 'sqlite://'
+        }
+        db = fsa.SQLAlchemy(app, engine_options=dict(echo=True))
+        db.create_all()
+
+        self.assertTrue(db.engine.echo)
+
 
 class DefaultQueryClassTestCase(unittest.TestCase):
     def test_default_query_class(self):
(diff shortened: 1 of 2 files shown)


test_sqlalchemy.py
```diff
@@ -503,6 +503,16 @@ class FooBoundModel(AbstractFooBoundModel):
         self.assertEqual(len(metadata.tables), 1)
         self.assertTrue('foo_bound_model' in metadata.tables)
 
+    def test_extended_engine_options(self):
+        app = flask.Flask(__name__)
+        app.config['SQLALCHEMY_BINDS'] = {
+            'foo': 'sqlite://'
+        }
+        db = fsa.SQLAlchemy(app, engine_options=dict(echo=True))
```
ThiefMaster: Probably not the best example for a unit test considering this one can be set through the flask config (`SQLALCHEMY_ECHO`)

rsyring:
For what it's worth, I think the approach I documented in [this comment](https://github.com/mitsuhiko/flask-sqlalchemy/issues/166#issuecomment-281398183) is a better approach that this PR as it permits the engine options to be set in the FLask configuration.

The suggestion:

```python
            options = {'convert_unicode': True}
            self._sa.apply_pool_defaults(self._app, options)
            self._sa.apply_driver_hacks(self._app, info, options)
            if echo:
                options['echo'] = echo

            # next two lines are new
            config_engine_opts = self._app.config.get('SQLALCHEMY_ENGINE_OPTS', {})
            options.update(config_engine_opts)

            self._engine = rv = sqlalchemy.create_engine(info, **options)
```

And it's only two lines difference.

asteinlein:
👍 to having this in Flask configuration.

rsyring:

keithjstone:
Huge thanks for the fix! We had found a workaround locally so we could continue to use the published versions. Will be looking to update!

PR #488, opened 2017-03-23
Add Windows example for SQLite URI

greyli:


ThiefMaster:
Is using backslashes really the preferred way there? I think usually people tend to use forward slashes in code even on windows (since all the windows APIs actually support them internally). 

When using backslashes I'd mention that people should use a `r'...'` string in their config

greyli:
Thanks, I added both choices.

greyli:
Why the edit in docs will fail to pass [test_session_scoping_changing()](https://travis-ci.org/mitsuhiko/flask-sqlalchemy/jobs/214185897#L207)?

spyoungtech:
@ThiefMaster 

The reason being that the connection URI `'sqlite:////absolute/path/to/foo.db'` Will not work as expected on Windows.

Instead, to get the expected effect, you need the URI to be
```python
'sqlite:///C:\\absolute\\path\\to\\foo.db'
```
or
```python
r'sqlite:///C:\absolute\path\to\foo.db'
```

I'm not sure if forward slashes will work, but I think in either case you need the drive letter for absolute path on Windows for an SQLite URI. Please check #488 again, which links the SQLAlchemy docs, which I believe is most appropriate for the Flask-SQLAlchemy counterpart docs.

greyli:
I think the drive letter is not necessary, since people may store the database file in other drive. Besides, the drive letter was included in the absolute path. 

I do add the raw string variant example in this [commit](https://github.com/mitsuhiko/flask-sqlalchemy/pull/488/commits/25fdd07b798293563cb1af78337091fa150e7d4a), but It fail to pass the check.

PR #491, opened 2017-03-30
init_app error

sonus21:
This fixes init_app function bug. Function init_app was not populating self.app which results in SQL query error, `RuntimeError: application not registered on db instance and no application bound to current context`


ThiefMaster:
`self.app` is intentionally left None when using init_app since you can use the extension on more than one Flask app. You are *supposed to* be inside an app context in this case whenever you want to use the extension.

PR #493, opened 2017-04-19
make calling_context more accurate

blurrcat:
As detailed in https://github.com/mgood/flask-debugtoolbar/issues/96, the current logic for generating the calling context often fails unnecessarily. 

Specifically, when a project is structured like the following,
```
project
├── project
│   ├── app.py
│   └── views.py
└── setup.py
```
and the flask application is defined by `app = Flask(__name__)` in `app.py`, the calling context will be "unknown" if the query is issued from somewhere other than `app.py`. This is because now `app_path` is `project.app` but the relevant frame that issues the query might be `project.views`.

The updated logic matches the frame with the top level package of the `app_path`.

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -120,9 +120,11 @@ class _DebugQueryTuple(tuple):
 
 def _calling_context(app_path):
     frm = sys._getframe(1)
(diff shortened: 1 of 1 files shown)


davidism:
Fixed in #1087

Issue #469, opened 2017-02-27
Add Windows example for sqlite URI configuration.


spyoungtech:
I think it would be helpful to include a windows example for the connection URI.

For example, the [SQLAlchemy engine configuration documentation](http://docs.sqlalchemy.org/en/latest/core/engines.html#sqlite) currently reads as follows:

```python
#Unix/Mac - 4 initial slashes in total
engine = create_engine('sqlite:////absolute/path/to/foo.db')
#Windows
engine = create_engine('sqlite:///C:\\path\\to\\foo.db')
#Windows alternative using raw string
engine = create_engine(r'sqlite:///C:\path\to\foo.db')
```

I think it would be helpful to add this or similar notes to the  [configuration](http://flask-sqlalchemy.pocoo.org/2.1/config/) docs and perhaps the [quickstart guide](http://flask-sqlalchemy.pocoo.org/2.1/quickstart/#a-minimal-application) as well.

A student of mine brought up a potentially misleading remark in the [current documentation](http://flask-sqlalchemy.pocoo.org/2.1/config/#connection-uri-format) that reads as follows (emphasis added is my own).

> SQLite **(note the four leading slashes)**:
> `sqlite:////absolute/path/to/foo.db`

The student took this to mean that the four leading slashes are absolutely critical and proceeded to try using the URI **`sqlite:////C:\Users\Student\path\to\foo.db`** When they should have only used 3 leading slashes, then the absolute path to the URL, ie `sqlite:///C:\path\to\foo.db`. Other students using Windows remarked having the same issue, and some resolved it by finding the mentioned SQLAlchemy docs.

PR #495, opened 2017-05-07
Add Windows example for sqlite URI configuration.

spyoungtech:
Addresses #469 -- adding windows example for URI configuration.

The current docs say
```
SQLite (note the four leading slashes):
    sqlite:////absolute/path/to/foo.db
```
However, this can lead to unexpected results on Windows. This PR modifies the notes and adds windows examples with notes. Much like the counterpart [SQLAlchemy docs](http://docs.sqlalchemy.org/en/latest/core/engines.html#sqlite)

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -115,10 +115,11 @@ SQLite (note that platform path conventions apply)::
 
     #Unix/Mac (note the four leading slashes)
     sqlite:////absolute/path/to/foo.db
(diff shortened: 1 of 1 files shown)


spyoungtech:
Further, it may be worthwhile to apply similar changes to the quickstart quide. This PR does NOT change the quickstart guide.

Issue #363, opened 2015-12-11
Use pytest, test dependency versions


davidism:
Flask converted, we should too.  They also test at the minimal, current, and development versions of their dependencies.  I don't think we need to go development, but we definitely need to test at SQLAlchemy==0.8.
- [x] convert to pytest
- [x] split tests into multiple files
- [x] add tests at minimal dependencies to tox
- [x] adjust travis to use new test targets

davidism:
added tox/travis envs in #388

thrawny:
Are looking for help with converting to pytest? I would like to help out.

davidism:
@thrawny yes, that would be a big help! Some things I'd like to do:

* split the tests into different files instead of one big one
* use flat functions instead of classes
* use fixtures where appropriate

thrawny:
@davidism Great! Thanks for the input. Will start working on it then.

PR #500, opened 2017-05-15
Convert tests to pytest

thrawny:
For #363 

Removed `from __future__ import with_statement` when converting. As far as I can see, python 2.5 is not supported.

Ran tests using tox locally and `test_session_scoping_changing` sometimes fails with `py27-lowest`. Maybe someone can elaborate on this.

In `test_sessions.py` I set `flask.testing` to `False` since that is how it was in the original tests. Not sure if this is on purpose or not. If not then the overridden app fixture in that file is not needed.

Otherwise straight forward conversion using fixtures and so on.


ThiefMaster:
Usually `tests` should not be a package.

thrawny:
@ThiefMaster Was mainly due to wanting to import `make_todo_model`from somewhere. Should I make a separate file for it? Importing from `conftest`is ugly imo.

ThiefMaster:
If it's a fixture in conftest.py you do not need to import it at all

davidism:
You basically already have the fixture, inline `make_todo_model` into the `Todo` fixture.

davidism:
For that test that constructs the model separately, just make a simple model in the test, you're not required to reuse that todo model everywhere.

thrawny:
Yeah great, will fix that then.

thrawny:
Removed `__init__.py` from `tests` and inlined the `Todo` model.

Had to change the `app` fixture a little bit to make other tests work. It get the current module name from the pytest request object and sets that as the app name.

thrawny:
@davidism Had a change to look at the changes I made?

PR #502, opened 2017-06-01
warning about naming your table User

michael-wise:
Put in a warning to remind inexperienced users that using the table name User can cause confusion, as this table name is reserved in several databases (notably, Postgres). Although these docs use sqlite3, new users may be plugging in different databases and should not be worried with the table name. Ideally, the table/class name in docs should be unanimously changed to Users.


davidism:
SQLAlchemy already quotes it properly.

michael-wise:
Ah, It was actually in reference to raw SQL--either through execute() or a DB CLI. SQLA newcomers (possibly avoiding SQL queries) might be the ones who need to stumble across that caution).

PR #503, opened 2017-06-04
allow module overload while engine creation

easyguy45:
In some scenarios I found it useful being able to specify the sqlalchemy module directly via bypassing. I use it for example in combination with a pysqlcipher3 binding.

This commit makes this possible while using the flask-sqlalchemy bindings.


davidism:
I don't want to add any more individual configuration options. Closing in favor of something like #487 that just allows configuring all the options.

easyguy45:
He davidism, I like #487  myself very much. Any timeline for integrating this commit / the next release?

PR #506, opened 2017-06-13
Fixed an error:

yiyuhao:
code and appearance:

        db = SQLAlchemy()
        # Now db.app is None

        # at __init__ in class SQLAlchemy(),  self.app = None
        db.init_app(app)

        # when exe at:
        #     app = self.get_app(app) in _execute_for_all_tables()
        #     self.app is still None
        #     so mistakenly raise a raise RuntimeError(
        #                                                  'application not registered on db instance and no application'
        #                                                  'bound to current context')
        db.create_all()


ThiefMaster:
`self.app` is supposed to be `None` if you use `init_app` since you might have more than one app.

PR #517, opened 2017-07-12
Change returning app order in `get_app` of connectors

### LINKED ISSUE #490: `get_app()` in connectors return uncertainty `app`
`SQLAlchemyState.connectors` hold the app get from `get_app()`. When connectors are created under flask context, theirs `_app` is `current_app`. Otherwise, theirs app is `SQLAlchemy.app`. Connectors created under the second scene work fine without flask context, while those created under the first scene raise exceptions.

Also if connectors have created under the second scene, they are not changed along with the flask context. In this case, the purpose of making `current_app` is not achieved.

lifei:
This fix #490

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -947,11 +947,11 @@ class SQLAlchemy(object):
         if reference_app is not None:
             return reference_app
 
-        if current_app:
-            return current_app
-
         if self.app is not None:
             return self.app
+
+        if current_app:
+            return current_app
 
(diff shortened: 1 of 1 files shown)


ThiefMaster:
See https://github.com/mitsuhiko/flask-sqlalchemy/issues/490#issuecomment-314720313

FYI: Your commit messages is excessively long (first line should be <51 chars) and contains typos.

davidism:
Fixed in #547. The order wasn't the issue, needed to return `current_app._get_current_object()`.

PR #521, opened 2017-07-23
Measure test coverage

singingwolfboy:
And send reports to [codecov.io](https://codecov.io/gh/mitsuhiko/flask-sqlalchemy)


davidism:
Thanks, I implemented this in #543 with a bit more based on Flask's config.

PR #333, opened 2015-08-20
Added a way to specify engine pool class in config

tmeryu:
Fixes mitsuhiko/flask-sqlalchemy#266

I didn't like the pool_size == 0 check (https://github.com/mitsuhiko/flask-sqlalchemy/pull/215) because it's a valid option given the SQLAlchemy documentation.

PR #525, opened 2017-08-08
Allow for arbitrary Pool configuration

mistercrunch:
I was looking for more flexibility around pools and this hook should
allow for users to pass whatever pool object they want to use.


davidism:
Is this a duplicate / related to #333?

mistercrunch:
This PR here may be more flexible as you can instantiate whatever pool object you want with whatever parameter.

mistercrunch:
@davidism can we get this through? I think this PR is the most flexible as it allows users to pass whatever pool configuration they may want

rsyring:

PR #526, opened 2017-08-10
fix: maybe we should not modify `app.config`

hidva:
```sh
$ cat web.config
SQLALCHEMY_DATABASE_URI = 'mysql+pymysql://******?charset=utf8'
```

```py
app = flask.Flask(__name__)
app.config.from_pyfile('web.config')
db = SQLAlchemy(app)
sqlengine = sqlalchemy.create_engine(app.config['SQLALCHEMY_DATABASE_URI'],
    pool_recycle=app.config.get('SQLALCHEMY_POOL_RECYCLE', 7200),  # pool_recycle is None! BAD!
    max_overflow=app.config.get('SQLALCHEMY_MAX_OVERFLOW', -1)) # max_overflow is None! BAD!
conn = sqlengine.connect()  # TERRIBLE!
```


rsyring:
There would be a decent amount of work to change this behavior and I'm not convinced there is a lot of additional value.  Modifying the app config is a common practice in the Flask world.

Issue #529, opened 2017-08-20
Missing file: test_sqlalchemy.py


xliiv:
`Readme`, `make` reference file:
test_sqlalchemy.py
There is no such file.

ThiefMaster:
The new way of running tests is with `pytest`. Feel free to send a PR to update the docs! I would use this opportunity to replace `py.test` with the non-deprecated version `pytest` in this line: https://github.com/mitsuhiko/flask-sqlalchemy/blob/master/tox.ini#L5

PR #530, opened 2017-08-20
Add default __repr__ for Model class

### LINKED ISSUE #528: [question] Add default __repr__ for Model?
Could be possible to make `Model` (or any variant of it) have default `__repr__`?
Like this:
```
class Model(object):
    """Base class for SQLAlchemy declarative base model.

    To define models, subclass :attr:`db.Model <SQLAlchemy.Model>`, not this class.
    To customize ``db.Model``, subclass this and pass it as ``model_class`` to :func:`SQLAlchemy`.
    """

    #: Query class used by :attr:`query`.
    #: Defaults to :class:`SQLAlchemy.Query`, which defaults to :class:`BaseQuery`.
    query_class = None

    #: Convenience property to query the database for instances of this model using the current session.
    #: Equivalent to ``db.session.query(Model)`` unless :attr:`query_class` has been changed.
    query = None

    _cached_tablename = None

    @declared_attr
    def __tablename__(cls):
        if (
            '_cached_tablename' not in cls.__dict__ and
            _should_set_tablename(cls)
        ):
            cls._cached_tablename = camel_to_snake_case(cls.__name__)

        return cls._cached_tablename
    # HERE'S THE ADD
    def __repr__(self):
        return '<{} {}>'.format(self.__class__.__name__, self.id)
```

I know not every table has column id, but thing is that I see a lot of this boilerplate in flask models.
So it could be handled by this lib somehow?

xliiv:
Add default `__repr__` for `Model` to reduce boilerplate

Fixes #528

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -663,9 +663,8 @@ class Model(object):
     query = None
 
     def __repr__(self):
-        pks = inspect(self).identity_key[1]
-        pretty_pk = pks[0] if len(pks) == 1 else ', '.join(map(str, pks))
-        return '<{0}({1})>'.format(type(self).__name__, pretty_pk)
+        pk = ', '.join(str(value) for value in inspect(self).identity)
+        return '<{0} {1}>'.format(self.__class__.__name__, pk)
 
 
 class SQLAlchemy(object):
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -30,7 +30,7 @@ from sqlalchemy.ext.declarative import DeclarativeMeta, declarative_base, \
 from sqlalchemy.orm.exc import UnmappedClassError
 from sqlalchemy.orm.session import Session as SessionBase
 
-from ._compat import itervalues, string_types, xrange
+from ._compat import itervalues, string_types, xrange, to_str
 
 __version__ = '2.2.1'
 
@@ -663,8 +663,8 @@ class Model(object):
     query = None
 
     def __repr__(self):
-        pk = ', '.join(str(value) for value in inspect(self).identity)
-        return '<{0} {1}>'.format(self.__class__.__name__, pk)
+        pk = ', '.join(to_str(value) for value in inspect(self).identity)
+        return '<{0} {1}>'.format(type(self).__name__, pk)
 
 
 class SQLAlchemy(object):
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -662,6 +662,11 @@ class Model(object):
     #: Equivalent to ``db.session.query(Model)`` unless :attr:`query_class` has been changed.
     query = None
 
+    def __repr__(self):
+        pks = inspect(self).identity_key[1]
+        pretty_pk = pks[0] if len(pks) == 1 else ', '.join(map(str, pks))
+        return '<{0}({1})>'.format(type(self).__name__, pretty_pk)
+
 
 class SQLAlchemy(object):
     """This class is used to control the SQLAlchemy integration to one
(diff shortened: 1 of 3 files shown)


xliiv:
I'd like to add unit-tests but perhaps there is an issue?
#529

flask_sqlalchemy/__init__.py
```diff
@@ -663,9 +663,8 @@ class Model(object):
     query = None
 
     def __repr__(self):
-        pks = inspect(self).identity_key[1]
-        pretty_pk = pks[0] if len(pks) == 1 else ', '.join(map(str, pks))
-        return '<{0}({1})>'.format(type(self).__name__, pretty_pk)
+        pk = ', '.join(str(value) for value in inspect(self).identity)
+        return '<{0} {1}>'.format(self.__class__.__name__, pk)
```
ThiefMaster: AFAIK `type(self)` is the preferred way to get the class (i remember asking this on SO a long time ago)

flask_sqlalchemy/__init__.py
```diff
@@ -663,9 +663,8 @@ class Model(object):
     query = None
 
     def __repr__(self):
-        pks = inspect(self).identity_key[1]
-        pretty_pk = pks[0] if len(pks) == 1 else ', '.join(map(str, pks))
-        return '<{0}({1})>'.format(type(self).__name__, pretty_pk)
+        pk = ', '.join(str(value) for value in inspect(self).identity)
```
ThiefMaster: let's test the `str(value)` with a PK that contains non-ascii characters. such a PK is pretty 💩 but you never know...

flask_sqlalchemy/__init__.py
davidism: Yep, first result for "type(self) vs self.__class__": https://stackoverflow.com/questions/10386166/python-self-class-vs-typeself 😄 They're equivalent, `type` just means one less function call.

tests/test_model_class.py
```diff
@@ -11,3 +11,26 @@ class SomeModel(db.Model):
...
+        name = db.Column(db.String, primary_key=True)
+
+    class Report(db.Model):
+        id = db.Column(db.Integer, primary_key=True, autoincrement=False)
+        user_name = db.Column(db.ForeignKey(User.name), primary_key=True)
+
+    db.create_all()
+
+    u = User(name='test')
+    db.session.add(u)
```
ThiefMaster: I think create_all, add and flush aren't needed here since you don't need a database roundtrip to access the attributes you just set

davidism: Unfortunately, `identity` isn't populated until the object is persisted.

ThiefMaster: ah, good point

flask_sqlalchemy/__init__.py
davidism: Yep, that fails. Have a suggestion how to rewrite this for Unicode on Py2? Will need to dig into Flask's compat stuff, maybe there's something there.

ThiefMaster: `repr()` instead of `str()`? slightly uglier in py2 because of the `u` prefix, but pretty much all PKs are numeric anyway

xliiv:
what's going on here? :)
should i finish this PR, or someone of you has taken it?

davidism:
I've rebased against next release branch and am doing some further fixes. The test wasn't testing anything before, and we need to handle Unicode keys. I'll finish this when I get back from lunch.

xliiv:
+1 for unicode, this `str(..)` was a bit bold
Could you explain why test was broken?

davidism:
`assert instance.__repr__, "<SomeModel({})>".format(instance.pk)` is really `assert instance.__repr__`, which is always true.

xliiv:
oh crap :) i should have put
`assert instance.__repr__ == "<SomeModel({})>".format(instance.pk)`

ThiefMaster:
nope, `repr(instance)` - there's no reason to call `__repr__` directly (and to call it you'd have needed `()` there anyway ;p)

xliiv:
yup, right `__repr__` it's method but in this way.. it at least have failed :)

PR #535, opened 2017-08-29
Added logic to abort with 404 on first page

bibumathew:
Fix for paginate to abort with 404 if no items are found

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -427,12 +427,17 @@ class BaseQuery(orm.Query):
             abort(404)
         return rv
 
-    def paginate(self, page=None, per_page=None, error_out=True):
+    def paginate(self, page=None, per_page=None, error_out=True,
+                 first_page_404=False):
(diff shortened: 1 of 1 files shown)


ThiefMaster:
404ing on the first page if it's empty doesn't sound like behavior people would expect. It also contradicts the documented behavior.

bibumathew:
Would it be a good idea to people control the desired behavior based on params
```
 def paginate(self, page=None, per_page=None, error_out=True,
                 first_page_404=False):
```

CoburnJoe:
Thanks so much for your effort @bibumathew. 

Given the age of this, the current conflicts and the lack of GitHub activity, I'd like to suggest closing this unless anyone can make a compelling argument to keep it?

I think it adds an unusual behavior such that the first page now has a default behavior to always abort if no results and anyone wishing to get the old behavior has to pass a new arg wherever they use paginate. At the very least, this should maintain backward compatibility by defaulting to False (but as @ThiefMaster says, it's not a great solution).

PR #540, opened 2017-09-16
Support additional SQLAlchemy parameters.

yangbaepark:
I found some SQLAlchemy parameters are helpful when we have unreliable network connectivity between applications and database.
* `creator` helps to create custom DBAPI connections.
* `echo_pool` will log all check-ins/outs of the pool.
* `pool_pre_ping` tests connections for linveness upon each checkout.
* `pool_reset_on_return` allows to override default behavior upon return.

By default, all those options are turned off (or following the default behavior in SQLAlchemy) for backward compatibility.

For more information about the parameters, please refer to the following official SQLAlchemy documentation:
http://docs.sqlalchemy.org/en/latest/core/engines.html#engine-creation-api

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -14,56 +14,73 @@ A list of configuration keys currently understood by the extension:
 
 .. tabularcolumns:: |p{6.5cm}|p{8.5cm}|
 
-================================== =========================================
-``SQLALCHEMY_DATABASE_URI``        The database URI that should be used for
-                                   the connection.  Examples:
+=================================== =========================================
+``SQLALCHEMY_DATABASE_URI``         The database URI that should be used for
+                                    the connection.  Examples:
 
-                                   - ``sqlite:////tmp/test.db``
-                                   - ``mysql://username:password@server/db``
-``SQLALCHEMY_BINDS``               A dictionary that maps bind keys to
-                                   SQLAlchemy connection URIs.  For more
-                                   information about binds see :ref:`binds`.
-``SQLALCHEMY_ECHO``                If set to `True` SQLAlchemy will log all
-                                   the statements issued to stderr which can
-                                   be useful for debugging.
-``SQLALCHEMY_RECORD_QUERIES``      Can be used to explicitly disable or
-                                   enable query recording.  Query recording
-                                   automatically happens in debug or testing
-                                   mode.  See :func:`get_debug_queries` for
-                                   more information.
-``SQLALCHEMY_NATIVE_UNICODE``      Can be used to explicitly disable native
-                                   unicode support.  This is required for
-                                   some database adapters (like PostgreSQL
-                                   on some Ubuntu versions) when used with
-                                   improper database defaults that specify
-                                   encoding-less databases.
-``SQLALCHEMY_POOL_SIZE``           The size of the database pool.  Defaults
-                                   to the engine's default (usually 5)
-``SQLALCHEMY_POOL_TIMEOUT``        Specifies the connection timeout for the
-                                   pool.  Defaults to 10.
-``SQLALCHEMY_POOL_RECYCLE``        Number of seconds after which a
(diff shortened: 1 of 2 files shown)


gilbsgilbs:
I'm also willing to see those options in flask-sqlalchemy. They would need to be documented though.

redixin:
Some of these options are available only in sqlalchemy 1.2.0b3 (pre release)

gilbsgilbs:
@redixin Yes, but it doesn't break support for previous versions anyways. It should just be documented.

redixin:
@gilbsgilbs 
Need to update sqlalchemy first
https://github.com/mitsuhiko/flask-sqlalchemy/blob/master/setup.py#L34

gilbsgilbs:
@redixin I don't see your point, could you elaborate? This selector is compatible with sqlalchemy 1.2+.

johanols:
Looking closed and open PRs it seems that this is a highly requested feature. I for one would love to be able to set `pool_pre_ping` without having to sub-class `SQLAlchemy`. Reading through the comments, it assess these are the still open issues with this PR:

* The added parameters needs to be documented.
* The version pinning of the `sqlalchemy` package needs to be adjusted.

Is there a possibility to get this fixed and merged within the scope of this PR?

redixin:
@gilbsgilbs 

> @redixin I don't see your point, could you elaborate? This selector is compatible with sqlalchemy 1.2+.

Version 1.2 wasn't released at that moment. It is released now, but we still have 'SQLAlchemy>=0.8.0' in setup.py

Need to update setup.py as well.

johanols:
@gilbsgilbs Thank you for your quick respond!

So making this PR merge ready would be really simple. Unfortunately, the original author doesn't seem to be very active. Would it in that case make sense to abandon this PR in favor of a new one?

I could arrange that if desired.

yangbaepark:
Hello Olsson and Gilbert,

I am sorry for the late response. I will update PR.

Regards,
Yangbae

davidism:
It doesn't look like you need to update anything right now, I'm not sure why users have commented saying you need to do so.

gilbsgilbs:
> Version 1.2 wasn't released at that moment. It is released now, but we still have 'SQLAlchemy>=0.8.0' in setup.py

Changing this selector would be a breaking change. There is no reason to prevent people from using this library with SQLAlchemy 1.1 and downwards, the library will keep working as before. We just need to document that the `POOL_PRE_PING` option will only work starting from SQLAlchemy 1.2+ (and document all the options that are added in this PR).

davidism:
Just a heads up though, this probably won't get merged as I'd rather support a more general way to pass arguments to the pool, engine, and session rather than adding individual config keys. I'm leaving it open mainly because we'll need to organize all the related PRs and issues at some point.

yangbaepark:
> 
> 
> Just a heads up though, this probably won't get merged as I'd rather support a more general way to pass arguments to the pool, engine, and session rather than adding individual config keys. I'm leaving it open mainly because we'll need to organize all the related PRs and issues at some point.
> 
> Thanks for your work though, I don't mean to say it's not appreciated, just that it points to a larger goal we have.

Sure, that makes sense. I agree it won't be sustainable to keep adding keys whenever SQLAlchemy introduces new parameters.

rsyring:
@davidism is this approach sufficient for you?

https://github.com/pallets/flask-sqlalchemy/issues/166#issuecomment-281398183

If so, a PR should be trivial.

davidism:
Sure, that could work, I'd like it to pull from a class/instance init attribute of defaults as well so it's not only configurable via config (:thinking:).

rsyring:
@davidism I can do that.

rsyring:
Thank you for your contribution and sorry this took so long. We opted to go in a different direction but it will solve your use case. Please see: #684

PR #476, opened 2017-02-27
Add option to disable auto table name generation

ThiefMaster:

PR #541, opened 2017-09-18
rewrite tablename generation again

### LINKED ISSUE #470: Broken compatibility on generation of joined table inheritance with abstract models
First of all many thanks for releasing 2.2!

Here is the test case for my problem:

```python
import flask
import flask_sqlalchemy as fsa
from sqlalchemy.ext.declarative import declared_attr


app = flask.Flask(__name__)
db = fsa.SQLAlchemy(app)


class CustomBaseForJoinedInheritance(db.Model):
    __abstract__ = True

    @declared_attr
    def __tablename__(cls):
        return cls.__name__.lower()

    @declared_attr
    def id(cls):
        return db.Column('id', db.Integer, primary_key=True)


class Employee(CustomBaseForJoinedInheritance):
    name = db.Column(db.String(50))
    type = db.Column(db.String(50))

    __mapper_args__ = {
        'polymorphic_identity': 'employee',
        'polymorphic_on': type
    }


class Engineer(Employee):
    id = db.Column(db.Integer, db.ForeignKey('employee.id'), primary_key=True)
    engineer_name = db.Column(db.String(30))

    __mapper_args__ = {
        'polymorphic_identity': 'engineer',
    }


class BadEmployee(CustomBaseForJoinedInheritance):
    name = db.Column(db.String(50))
    type = db.Column(db.String(50))

    __mapper_args__ = {
        'polymorphic_identity': 'employee',
        'polymorphic_on': type
    }


class BadEngineer(BadEmployee):
    id = db.Column(db.Integer, db.ForeignKey('bademployee.id'), primary_key=True)
    engineer_name = db.Column(db.String(30))

    __mapper_args__ = {
        'polymorphic_identity': 'engineer',
    }


assert Employee.__tablename__ == 'employee'
assert Engineer.__tablename__ == 'engineer'
assert BadEmployee.__tablename__ == 'bademployee'
assert BadEngineer.__tablename__ == 'badengineer'

# works until here
db.create_all()
assert True
```

It's a bit long but please bear with me; all equality assertions pass greatly but calling `db.create_all()` raises following exception: 

```python
  File "test_case.py", line 51, in <module>
    class BadEngineer(BadEmployee):
  File "/usr/lib64/python3.5/site-packages/flask_sqlalchemy/__init__.py", line 602, in __init__
    DeclarativeMeta.__init__(self, name, bases, d)
  File "/usr/lib64/python3.5/site-packages/sqlalchemy/ext/declarative/api.py", line 64, in __init__
    _as_declarative(cls, classname, cls.__dict__)
  File "/usr/lib64/python3.5/site-packages/sqlalchemy/ext/declarative/base.py", line 88, in _as_declarative
    _MapperConfig.setup_mapping(cls, classname, dict_)
  File "/usr/lib64/python3.5/site-packages/sqlalchemy/ext/declarative/base.py", line 103, in setup_mapping
    cfg_cls(cls_, classname, dict_)
  File "/usr/lib64/python3.5/site-packages/sqlalchemy/ext/declarative/base.py", line 135, in __init__
    self._early_mapping()
  File "/usr/lib64/python3.5/site-packages/sqlalchemy/ext/declarative/base.py", line 138, in _early_mapping
    self.map()
  File "/usr/lib64/python3.5/site-packages/sqlalchemy/ext/declarative/base.py", line 530, in map
    **self.mapper_args
  File "<string>", line 2, in mapper
  File "/usr/lib64/python3.5/site-packages/sqlalchemy/orm/mapper.py", line 671, in __init__
    self._configure_inheritance()
  File "/usr/lib64/python3.5/site-packages/sqlalchemy/orm/mapper.py", line 978, in _configure_inheritance
    self.local_table)
  File "<string>", line 2, in join_condition
  File "/usr/lib64/python3.5/site-packages/sqlalchemy/sql/selectable.py", line 976, in _join_condition
    (a.description, b.description, hint))
sqlalchemy.exc.NoForeignKeysError: Can't find any foreign key relationships between 'bad_employee' and 'bad_engineer'.
``` 
Interestingly we have: `bad_employee` and `bad_engineer`, not `bademployee` and `badengineer`. It was working in 2.1, stopped in 2.2 (probably after #467). Do you plan to support such test case?

### LINKED ISSUE #473: InvalidRequestError: Class does not have a __table__ or __tablename__ specified and does not inherit from an existing table-mapped class.
Hi,

First of all congrats on the new release! This will allow us to go back to a pinned version of flask-sqlalchemy instead of the master branch 👍 

On the other hand, the change #467 now generates this stack-trace in the [Flask-AppBuilder](https://github.com/dpgaspar/Flask-AppBuilder/) project which I would need some help investigating to be able to use release 2.2 directly:
```
12:48:03   File "/src/<module_name>/models.py", line 10, in <module>
12:48:03     from flask_appbuilder.security.sqla.models import User
12:48:03   File "/usr/local/lib/python2.7/dist-packages/flask_appbuilder/security/sqla/models.py", line 12, in <module>
12:48:03     class Permission(Model):
12:48:03   File "/usr/local/lib/python2.7/dist-packages/flask_sqlalchemy/__init__.py", line 602, in __init__
12:48:03     DeclarativeMeta.__init__(self, name, bases, d)
12:48:03   File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/ext/declarative/api.py", line 55, in __init__
12:48:03     _as_declarative(cls, classname, cls.__dict__)
12:48:03   File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/ext/declarative/base.py", line 88, in _as_declarative
12:48:03     _MapperConfig.setup_mapping(cls, classname, dict_)
12:48:03   File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/ext/declarative/base.py", line 103, in setup_mapping
12:48:03     cfg_cls(cls_, classname, dict_)
12:48:03   File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/ext/declarative/base.py", line 133, in __init__
12:48:03     self._setup_inheritance()
12:48:03   File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/ext/declarative/base.py", line 429, in _setup_inheritance
12:48:03     "table-mapped class." % cls
12:48:03 InvalidRequestError: Class <class 'flask_appbuilder.security.sqla.models.Permission'> does not have a __table__ or __tablename__ specified and does not inherit from an existing table-mapped class.
```

The error message contradicts the [model declaration](https://github.com/dpgaspar/Flask-AppBuilder/blob/master/flask_appbuilder/security/sqla/models.py#L12-L18) as it does have a `__tablename` defined:
```python
class Permission(Model):
    __tablename__ = 'ab_permission'
    id = Column(Integer, Sequence('ab_permission_id_seq'), primary_key=True)
    name = Column(String(100), unique=True, nullable=False)

    def __repr__(self):
        return self.name
```

I understand that this change is related to the `@declared_attr` decorator but despite reading http://docs.sqlalchemy.org/en/latest/orm/extensions/declarative/api.html#api-reference I don't really know exactly how I should apply it to the model.

Thanks for your help

### LINKED ISSUE #478: `InvalidRequestError: Table … is already defined for this MetaData instance` (table inheritance on v2.2)
On Flask-SQLAlchemy 2.2 (Flask 0.12, SQLAlchemy 1.1.6, Python 3.5.3) I the following traceback that does not occur with Flask-SQLAlchemy 2.1:

```pytb
  File "application.py", line 19, in <module>
    class ChildA(Parent):
  File "/home/user/project/venv/lib/python3.5/site-packages/flask_sqlalchemy/__init__.py", line 602, in __init__
    DeclarativeMeta.__init__(self, name, bases, d)
  File "/home/user/project/venv/lib/python3.5/site-packages/sqlalchemy/ext/declarative/api.py", line 64, in __init__
    _as_declarative(cls, classname, cls.__dict__)
  File "/home/user/project/venv/lib/python3.5/site-packages/sqlalchemy/ext/declarative/base.py", line 88, in _as_declarative
    _MapperConfig.setup_mapping(cls, classname, dict_)
  File "/home/user/project/venv/lib/python3.5/site-packages/sqlalchemy/ext/declarative/base.py", line 103, in setup_mapping
    cfg_cls(cls_, classname, dict_)
  File "/home/user/project/venv/lib/python3.5/site-packages/sqlalchemy/ext/declarative/base.py", line 131, in __init__
    self._setup_table()
  File "/home/user/project/venv/lib/python3.5/site-packages/sqlalchemy/ext/declarative/base.py", line 395, in _setup_table
    **table_kw)
  File "/home/user/project/venv/lib/python3.5/site-packages/sqlalchemy/sql/schema.py", line 418, in __new__
    "existing Table object." % key)
sqlalchemy.exc.InvalidRequestError: Table 'items' is already defined for this MetaData instance.  Specify 'extend_existing=True' to redefine options and columns on an existing Table object.
```

Somewhat minimal code to reproduce:
```py
from flask import Flask


db = SQLAlchemy()


class Parent(db.Model):
    __tablename__ = 'items'
    __mapper_args__ = {
        'polymorphic_identity': 'parent',
        'polymorphic_on': 'type',
    }

    type = db.Column(db.Unicode(10), nullable=False)


class ChildA(Parent):
    __mapper_args__ = {
        'polymorphic_identity': 'child_a',
    }

    foo = db.Column(db.Integer)


class ChildB(Parent):
    __mapper_args__ = {
        'polymorphic_identity': 'child_b',
    }

    bar = db.Column(db.Integer)


app = Flask(__name__)
app.run()
```

Unless I've been doing that wrong the whole time and just got away with it by chance, I think this might be unintended behavior on Flask-SQLAlchemy's side.

### LINKED ISSUE #481: Problem with STI: __tablename__ automatically set (unwanted behaviour)
Hey, guys.

I'm using:

* Flask-SQLAlchemy==2.2;
* SQLAlchemy==1.1.5.

And have the problem, originally described in [this SO question](http://stackoverflow.com/questions/42563434/sqlalchemy-using-sti-how-to-automatically-substitute-type-field/42567228#42567228). My models setup looks like the following:

```python


db = SQLAlchemy()


class Item(db.Model):
    __tablename__ = 'item'

    info = db.Column(db.String(255))
    type = db.Column(db.String())

    __mapper_args__ = {
        'polymorphic_on': type
    }


class ItemA(Item):
    __mapper_args__ = {
        'polymorphic_identity': 'a'
    }


class ItemB(Item):
    __mapper_args__ = {
        'polymorphic_identity': 'b'
    }


class ItemC(Item):
    __mapper_args__ = {
        'polymorphic_identity': 'c'
    }

print(ItemB.query)
```

I want to use [Single Table Inheritance](http://docs.sqlalchemy.org/en/latest/orm/inheritance.html#single-table-inheritance) (STI). So expected output of the `print(ItemB.query)` statement is something like:

```SQL
SELECT item.id AS item_id, item.info AS item_info, item.type AS item_type FROM item WHERE item.type IN (?)
```

And at the same time I'm getting:

```SQL
SELECT item.id AS item_id, item.info AS item_info, item.type AS item_type FROM item
```

[@univerio](http://stackoverflow.com/users/341730/univerio) (@univerio, is it you? 😃 ) has pointed that Flask-SQLAlchemy automatically set `__tablename__` attribute. It means that my model setup, presented above, doesn't follow STI model.

**Is there any way to disable such behaviour by default?** From my POV that behaviour sounds like a wrong approach, having a developer (how doesn't know that fact) to spend hours to understand why something isn't working.

Thanks.

### LINKED ISSUE #492: Declarative backref rels not filtering by polymorphic identity
More info here:
https://bitbucket.org/zzzeek/sqlalchemy/issues/3960/declarative-backref-not-filtering-by

"the issue is somewhere in db.Model where their declarative base is interfering with declarative being able to set the "single inheritance" flag on the Event1 mapper."

### LINKED ISSUE #512: NoForeignKeysError with polymorphism and schemas
```python
from flask import Flask

app = Flask(__name__)
app.config['SQLALCHEMY_TRACK_MODIFICATIONS'] = False
app.config['SQLALCHEMY_DATABASE_URI'] = 'postgresql:///test'
db = SQLAlchemy(app)


class Item(db.Model):
    __tablename__ = 'items'
    __table_args__ = {'schema': 'foo'}  # this causes the error
    __mapper_args__ = {'polymorphic_on': 'type', 'polymorphic_identity': None}

    type = db.Column(db.Integer, nullable=False)
    parent_id = db.Column(db.ForeignKey('foo.items.id'), index=True, nullable=True)
    children = db.relationship('Item', backref=db.backref('parent', remote_side=[id]))


class SubItem(Item):
    __mapper_args__ = {
        'polymorphic_identity': 1
    }

```

traceback:

```pythontraceback
  File "flasksatest.py", line 21, in <module>
    class SubItem(Item):
  File "/home/adrian/dev/indico/env/lib/python2.7/site-packages/flask_sqlalchemy/__init__.py", line 602, in __init__
    DeclarativeMeta.__init__(self, name, bases, d)
  File "/home/adrian/dev/indico/env/lib/python2.7/site-packages/sqlalchemy/ext/declarative/api.py", line 64, in __init__
    _as_declarative(cls, classname, cls.__dict__)
  File "/home/adrian/dev/indico/env/lib/python2.7/site-packages/sqlalchemy/ext/declarative/base.py", line 88, in _as_declarative
    _MapperConfig.setup_mapping(cls, classname, dict_)
  File "/home/adrian/dev/indico/env/lib/python2.7/site-packages/sqlalchemy/ext/declarative/base.py", line 103, in setup_mapping
    cfg_cls(cls_, classname, dict_)
  File "/home/adrian/dev/indico/env/lib/python2.7/site-packages/sqlalchemy/ext/declarative/base.py", line 135, in __init__
    self._early_mapping()
  File "/home/adrian/dev/indico/env/lib/python2.7/site-packages/sqlalchemy/ext/declarative/base.py", line 138, in _early_mapping
    self.map()
  File "/home/adrian/dev/indico/env/lib/python2.7/site-packages/sqlalchemy/ext/declarative/base.py", line 534, in map
    **self.mapper_args
  File "<string>", line 2, in mapper
  File "/home/adrian/dev/indico/env/lib/python2.7/site-packages/sqlalchemy/orm/mapper.py", line 671, in __init__
    self._configure_inheritance()
  File "/home/adrian/dev/indico/env/lib/python2.7/site-packages/sqlalchemy/orm/mapper.py", line 978, in _configure_inheritance
    self.local_table)
  File "<string>", line 2, in join_condition
  File "/home/adrian/dev/indico/env/lib/python2.7/site-packages/sqlalchemy/sql/selectable.py", line 979, in _join_condition
    (a.description, b.description, hint))
sqlalchemy.exc.NoForeignKeysError: Can't find any foreign key relationships between 'items' and 'items'.
```

It works fine if I do not use a custom schema.

### LINKED ISSUE #533: table name is not auto-generated while the primary key is defined in __table_args__
Take these code as example:

    app = Flask(__name__)
    db = SQLAlchemy()
    db.init_app(app)
    
    
    class Plan(db.Model):
        __table_args__ = (
            db.PrimaryKeyConstraint('id', name='pk_plan'),
            db.UniqueConstraint('name', name='uq_plan')
         )
        id = db.Column(db.Integer)
        name = db.Column(db.String(50))
    
    with app.app_context():
        db.create_all()

When try to generate the table, an error is occurred.  The trace back information is something like this:

      File "model.py", line 13, in <module>
        class Plan(db.Model):
        File "/home/zero/study/python_env/for_cl/lib/python3.5/sitepackages/flask_sqlalchemy/__init__.py", line 603, in __init__
          DeclarativeMeta.__init__(self, name, bases, d)
        File "/home/zero/study/python_env/for_cl/lib/python3.5/sitepackages/sqlalchemy/ext/declarative/api.py", line 64, in __init__
          _as_declarative(cls, classname, cls.__dict__)
        File "/home/zero/study/python_env/for_cl/lib/python3.5/sitepackages/sqlalchemy/ext/declarative/base.py", line 88, in _as_declarative
           _MapperConfig.setup_mapping(cls, classname, dict_)
        File "/home/zero/study/python_env/for_cl/lib/python3.5/sitepackages/sqlalchemy/ext/declarative/base.py", line 103, in setup_mapping
          cfg_cls(cls_, classname, dict_)
        File "/home/zero/study/python_env/for_cl/lib/python3.5/sitepackages/sqlalchemy/ext/declarative/base.py", line 133, in __init__
          self._setup_inheritance()
        File "/home/zero/study/python_env/for_cl/lib/python3.5/sitepackages/sqlalchemy/ext/declarative/base.py", line 430, in _setup_inheritance
          "table-mapped class." % cls
    sqlalchemy.exc.InvalidRequestError: Class <class '__main__.Plan'> does not have a __table__ or __tablename__ specified and does not inherit from an existing table-mapped class.

May be it can be enhanced by support the table defined schema like this ?

davidism:
A tablename will be set for non-abstract classes unless a mixin defines one or a model or mixin defines a `declared_attr`.

`__table_cls__` executes at a time when all column and primary key constraints are known, and can decide not to set a table when the model looks like single-table inheritance. It will also unset a tablename in this case.

I haven't added tests for the issues yet, but this should eventually fix #470, fix #473, fix #478, fix #481, fix #492, fix #512, fix #533. 😣

CC @zzzeek, @ThiefMaster 

---

Previously:

~Goes back to performing all tablename behavior in `_BoundDeclarativeMeta`, instead of putting some in a `declared_attr`. It would be more accurate to diff this against 2.1.~

~If you define a primary key in a `declared_attr`, you have to mark that attr as a the primary key, as we no longer evaluate those attrs against base classes.~

~Detects primary key constraint in `__table_args__`.~

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -30,7 +30,7 @@ from sqlalchemy.ext.declarative import DeclarativeMeta, declarative_base, \
 from sqlalchemy.orm.exc import UnmappedClassError
 from sqlalchemy.orm.session import Session as SessionBase
 
-from ._compat import iteritems, itervalues, string_types, xrange
+from ._compat import itervalues, string_types, xrange
 
 __version__ = '2.2.1'
 
@@ -551,31 +551,39 @@ class _EngineConnector(object):
 
 
 def _should_set_tablename(cls):
-    """Traverse the model's MRO. If a primary key column is found before a
-    table or tablename, then a new tablename should be generated.
+    """Determine whether ``__tablename__`` should be automatically generated
+    for a model.
 
-    This supports:
+    * If no class in the MRO sets a name, one should be generated.
+    * If a declared attr is found, it should be used instead.
+    * If a name is found, it should be used if the class is a mixin, otherwise
+      one should be generated.
+    * Abstract models should not have one generated.
 
-    * Joined table inheritance without explicitly naming sub-models.
-    * Single table inheritance.
-    * Inheriting from mixins or abstract models.
-
-    :param cls: model to check
-    :return: True if tablename should be set
+    Later, :meth:`._BoundDeclarativeMeta.__table_cls__` will determine if the
+    model looks like single or joined-table inheritance. If no primary key is
+    found, the name will be unset.
     """
+    if (
+        cls.__dict__.get('__abstract__', False)
+        or not any(isinstance(b, DeclarativeMeta) for b in cls.__mro__[1:])
+    ):
+        return False
 
     for base in cls.__mro__:
-        d = base.__dict__
+        if '__tablename__' not in base.__dict__:
+            continue
 
-        if '__tablename__' in d or '__table__' in d:
+        if isinstance(base.__dict__['__tablename__'], declared_attr):
             return False
 
-        for name, obj in iteritems(d):
-            if isinstance(obj, declared_attr):
-                obj = getattr(cls, name)
+        return not (
+            base is cls
+            or base.__dict__.get('__abstract__', False)
+            or not isinstance(base, DeclarativeMeta)
+        )
 
-            if isinstance(obj, sqlalchemy.Column) and obj.primary_key:
-                return True
+    return True
 
 
 def camel_to_snake_case(name):
@@ -591,20 +599,36 @@ def camel_to_snake_case(name):
 
 
 class _BoundDeclarativeMeta(DeclarativeMeta):
-    def __new__(cls, name, bases, d):
-        # if tablename is set explicitly, move it to the cache attribute so
-        # that future subclasses still have auto behavior
-        if '__tablename__' in d:
-            d['_cached_tablename'] = d.pop('__tablename__')
+    def __init__(cls, name, bases, d):
+        if _should_set_tablename(cls):
+            cls.__tablename__ = camel_to_snake_case(cls.__name__)
 
-        return DeclarativeMeta.__new__(cls, name, bases, d)
+        bind_key = (
+            d.pop('__bind_key__', None)
+            or getattr(cls, '__bind_key__', None)
+        )
 
-    def __init__(self, name, bases, d):
-        bind_key = d.pop('__bind_key__', None) or getattr(self, '__bind_key__', None)
-        DeclarativeMeta.__init__(self, name, bases, d)
+        super(_BoundDeclarativeMeta, cls).__init__(name, bases, d)
 
-        if bind_key is not None and hasattr(self, '__table__'):
-            self.__table__.info['bind_key'] = bind_key
+        if bind_key is not None and hasattr(cls, '__table__'):
+            cls.__table__.info['bind_key'] = bind_key
+
+    def __table_cls__(cls, *args, **kwargs):
+        """This is called by SQLAlchemy during mapper setup. It determines the
+        final table object that the model will use.
+
+        If no primary key is found, that indicates single-table inheritance,
+        so no table will be created and ``__tablename__`` will be unset.
+        """
+        for arg in args:
+            if (
+                (isinstance(arg, sqlalchemy.Column) and arg.primary_key)
+                or isinstance(arg, sqlalchemy.PrimaryKeyConstraint)
+            ):
+                return sqlalchemy.Table(*args, **kwargs)
+
+        if '__tablename__' in cls.__dict__:
+            del cls.__tablename__
 
 
 def get_state(app):
@@ -637,18 +661,6 @@ class Model(object):
     #: Convenience property to query the database for instances of this model using the current session.
     #: Equivalent to ``db.session.query(Model)`` unless :attr:`query_class` has been changed.
     query = None
-
-    _cached_tablename = None
-
-    @declared_attr
-    def __tablename__(cls):
-        if (
-            '_cached_tablename' not in cls.__dict__ and
-            _should_set_tablename(cls)
-        ):
-            cls._cached_tablename = camel_to_snake_case(cls.__name__)
-
-        return cls._cached_tablename
 
 
 class SQLAlchemy(object):
--- a/tests/test_table_name.py
+++ b/tests/test_table_name.py
@@ -1,3 +1,5 @@
+import inspect
+
 from sqlalchemy.ext.declarative import declared_attr
 
 
@@ -25,6 +27,7 @@ def test_single_name(db):
     class Mallard(Duck):
         pass
 
+    assert '__tablename__' not in Mallard.__dict__
     assert Mallard.__tablename__ == 'duck'
 
 
@@ -39,8 +42,10 @@ def test_joined_name(db):
     assert Donald.__tablename__ == 'donald'
 
 
-def test_mixin_name(db):
-    """Primary key provided by mixin should still allow model to set tablename."""
+def test_mixin_id(db):
+    """Primary key provided by mixin should still allow model to set
+    tablename.
+    """
     class Base(object):
         id = db.Column(db.Integer, primary_key=True)
 
@@ -51,8 +56,33 @@ def test_mixin_name(db):
     assert Duck.__tablename__ == 'duck'
 
 
+def test_mixin_attr(db):
+    """A declared attr tablename will be used down multiple levels of
+    inheritance.
+    """
+    class Mixin(object):
+        @declared_attr
+        def __tablename__(cls):
+            return cls.__name__.upper()
+
+    class Bird(Mixin, db.Model):
+        id = db.Column(db.Integer, primary_key=True)
+
+    class Duck(Bird):
+        # object reference
+        id = db.Column(db.ForeignKey(Bird.id), primary_key=True)
+
+    class Mallard(Duck):
+        # string reference
+        id = db.Column(db.ForeignKey('DUCK.id'), primary_key=True)
+
+    assert Bird.__tablename__ == 'BIRD'
+    assert Duck.__tablename__ == 'DUCK'
+    assert Mallard.__tablename__ == 'MALLARD'
+
+
 def test_abstract_name(db):
-    """Abstract model should not set a name.  Subclass should set a name."""
+    """Abstract model should not set a name. Subclass should set a name."""
     class Base(db.Model):
         __abstract__ = True
         id = db.Column(db.Integer, primary_key=True)
@@ -60,19 +90,23 @@ def test_abstract_name(db):
     class Duck(Base):
         pass
 
-    assert Base.__tablename__ == 'base'
+    assert '__tablename__' not in Base.__dict__
     assert Duck.__tablename__ == 'duck'
 
 
 def test_complex_inheritance(db):
-    """Joined table inheritance, but the new primary key is provided by a mixin, not directly on the class."""
+    """Joined table inheritance, but the new primary key is provided by a
+    mixin, not directly on the class.
+    """
     class Duck(db.Model):
         id = db.Column(db.Integer, primary_key=True)
 
     class IdMixin(object):
         @declared_attr
         def id(cls):
-            return db.Column(db.Integer, db.ForeignKey(Duck.id), primary_key=True)
+            return db.Column(
+                db.Integer, db.ForeignKey(Duck.id), primary_key=True
+            )
 
     class RubberDuck(IdMixin, Duck):
         pass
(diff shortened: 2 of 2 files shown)


flask_sqlalchemy/__init__.py
```diff
@@ -551,31 +555,99 @@ def get_engine(self):
...
+    Abstract classes do not generate a name, although they may have set
+    or inherited a name elsewhere.
+
+    If a class defines a tablename or table, a new one will not be generated.
+    Otherwise, if the class defines a primary key, a new name will be
+    generated.
+
+    This does not execute
+    :class:`sqlalchemy.ext.declarative.declared_attr` properties. If such an
+    attribute defines the primary key, decorate it with
```
ThiefMaster: Is this also necessary when not using table name generation at all?

davidism: No, nothing else would need to evaluate attrs prematurely except the tablename generation.

davidism:
Currently having a discussion with @zzzeek [over here](https://bitbucket.org/zzzeek/sqlalchemy/issues/4082/marking-subclass-of-declared_attr-with).

The reason for `declared_pkey` is that we want to support the use case "if the user doesn't set a primary key, they want single-table inheritance and shouldn't set a tablename". Because of this we need a way to distinguish if a primary key is set or not, even with `declared_attr`. If we dropped this requirement, we would always assume joined-table inheritance, and so would always generate a tablename regardless of primary key. Users would indicate they want STI by setting `__tablename__ = None`, which is how things basically work in plain SQLAlchemy when using a declared tablename anyway.

ThiefMaster:
I still don't like the idea of breaking models that work perfectly fine in vanilla SQLAlchemy. Would it make sense to allow tablename generation only for more basic use cases? I think an application complex enough to use polymorphism should set explicit tablenames anyway.

Or we could assume a certain behavior (like you suggested) when tablename generation is enabled, but let people opt out (there's #476 but I guess it'd need some changes to work with the changes from this PR) if they set explicit table names anyway.

ThiefMaster:
So after reading your discussion with @zzzeek, maybe it'd be a good idea to decide on something on the SQLAlchemy side for this:

> If it were me, I'd simply not have it work this way. Single inheritance can be indicated by any other means, including a flag like `single_inheritance = True`, or `__tablename__ = None`, etc. 

Then Flask-SQLAlchemy could already require this while SQLAlchemy itself would only add such a breaking change in 1.3 (since I believe SA allows breaking changes in 1.x versions)

zzzeek:
I wasn't referring to SQLAlchemy having any kind of breaking change.  If you have a base class with a __tablename__() method, it can look at "cls" for any kinds of flags like "cls.is_single" or something like that.

davidism:
Sorry, I removed my comment because more stuff came up in the discussion that I wanted to consider first.

Regardless of what I end up with, I do plan on having an "opt-out" when setting up the extension, similar to Thief's PR.

vToMy:
I believe I suggested a solution for 2 of those issues (#470, #481) without a complete rewrite, that might fix the others as well, as essentially we have a core problem.

davidism:
Based on discussion with zzzeek, I've pushed a new version that uses the `__table_cls__` hook.

davidism:
Uploading the patch for attempt 1, since it might be a useful reference still. 
[1c4640efa8352a070e91fd59254045668ead0c46.txt](https://github.com/mitsuhiko/flask-sqlalchemy/files/1333356/1c4640efa8352a070e91fd59254045668ead0c46.txt)

davidism:
Travis failure is unrelated. Will have to look into updating Travis matrix after this.

Issue #384, opened 2016-03-25
Provide a max_per_page parameter for pagination


dmulter:
Right now the default for per_page is 20 unless otherwise specified. I also see that the code makes sure it is no larger than the total items.

To prevent a request asking for too many items at once, it would be convenient if Pagination could be initialized with a maximum items per page (max_per_page). Then the supplied per_page value could use the ceil() of total and max_per_page.

kalombos:
#542

PR #542, opened 2017-09-21
limited per_page param in paginate

kalombos:
Hey, in my application i used BaseQuery.paginate() method without any params and found that user can be able to use unlimited value for per_page param in http query. I think this param should be limited, here is the PR.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -12,7 +12,10 @@ In development
   by models. (`#551`_)
 - Raise the correct error when a model has a table name but no primary key.
   (`#556`_)
+- Allow specifying a ``max_per_page`` limit for pagination, to avoid users
+  specifying high values in the request args. (`#542`_)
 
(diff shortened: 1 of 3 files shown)


ThiefMaster:
Sorry, but that's a very bad idea.

First of all, it makes LOTS of sense to allow more than 20 items for some applications, so setting an arbitrary limit is not a good idea.

If you want to limit the amount of items per page, get the value in your code and pass it to the `per_page` argument after checking/limiting it.

kalombos:
I see. Then what about #384 What if i would add max_per_page param to paginate method? As i understood it should be None by default.

kalombos:
Anyway i added commit with max_per_page param. Hope it will be useful.

davidism:
This seems fine, I didn't consider the issue that users could specify a large per page when I added the automatic args.

PR #544, opened 2017-09-27
Specifying the poolclass argument manually.

georgechang0117:
In my case, our one of databases which is based on SQL Server would close every mid night.
By default, the poolclass was set, but the query would failed if the database was restarted. 
Because the connection was still in memory for efficient reuse, it thought the database was still active.


davidism:
Is this duplicate / related to #525?

davidism:
Also, looks like you haven't configured your git user information.

georgechang0117:
I think so. They all want to solve similar / relational problems as like mine.

georgechang0117:
BTW, I updated my git user info. Thanks.

rsyring:

PR #546, opened 2017-09-28
allow passing declarative_base as model_class

davidism:
The original purpose for this was to be able to disable automatic tablename generation, but it became more general.

Previously, `model_class` needed to be a basic class, which would be passed to `declarative_base`. Now, if it is already the result of `declarative_base`, it is used directly. This allows greater customization by allowing the user to specify whatever arguments they want to `declarative_base`, including `metaclass`.

`_BoundDeclarativeMeta` has been renamed `DefaultMeta` since the previous name wasn't completely accurate. The `__bind_key__` and `__tablename__` behaviors are split into mixin classes that are both used to define `DefaultMeta`. Creating a metaclass without the `__tablename__` behavior looks like:

```python
from flask_sqlalchemy.model import Model, BindMetaMixin
from sqlalchemy.ext.declarative import DeclarativeMeta, declarative_base

class NoNameMeta(BindMetaMixin, DeclarativeMeta):
    pass

db = SQLAlchemy(model_class=declarative_base(cls=Model, metaclass=NoNameMeta))
```

### DIFF
--- a/docs/customizing.rst
+++ b/docs/customizing.rst
@@ -6,75 +6,81 @@ Customizing
 ===========
 
 Flask-SQLAlchemy defines sensible defaults. However, sometimes customization is
-needed. Two major pieces to customize are the Model base class and the default
-Query class.
+needed. There are various ways to customize how the models are defined and
+interacted with.
 
-Both of these customizations are applied at the creation of the :class:`SQLAlchemy`
+These customizations are applied at the creation of the :class:`SQLAlchemy`
 object and extend to all models derived from its ``Model`` class.
+
 
 Model Class
 -----------
 
-Flask-SQLAlchemy allows defining a custom declarative base, just like SQLAlchemy,
-that all model classes should extend from. For example, if all models should have
-a custom ``__repr__`` method::
+SQLAlchemy models all inherit from a declarative base class. This is exposed
+as ``db.Model`` in Flask-SQLAlchemy, which all models extend. This can be
+customized by subclassing the default and passing the custom class to
+``model_class``.
 
-    from flask_sqlalchemy import Model  # this is the default declarative base
-    from flask_sqlalchemy import SQLAlchemy
-
-    class ReprBase(Model):
-         def __repr__(self):
-             return "<{0} id: {1}>".format(self.__class__.__name__, self.id)
-
-    db = SQLAlchemy(model_class=ReprBase)
-
-    class MyModel(db.Model):
-        ...
+The following example gives every model an integer primary key, or a foreign
+key for joined-table inheritance.
 
 .. note::
 
-        While not strictly necessary to inherit from :class:`flask_sqlalchemy.Model`
-        it is encouraged as future changes may cause incompatibility.
+    Integer primary keys for everything is not necessarily the best database
+    design (that's up to your project's requirements), this is only an example.
 
-.. note::
-
-        If behavior is needed in only some models, not all, a better strategy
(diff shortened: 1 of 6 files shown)


docs/customizing.rst
```diff
@@ -6,75 +6,101 @@ Customizing
...
+primary keys::
+
+    from flask_sqlalchemy import Model, SQLAlchemy
+    import sqlalchemy as sa
 
-    class ReprBase(Model):
-         def __repr__(self):
-             return "<{0} id: {1}>".format(self.__class__.__name__, self.id)
+    class EqModel(Model):
+        def __eq__(self, other):
```
ThiefMaster: is `__eq__` a good example? pretty sure sqlalchemy already has a decent `__eq__` implementation

docs/customizing.rst
```diff
@@ -6,75 +6,101 @@ Customizing
...
+            return sa.inspect(self).identity == sa.inspect(other).identity
 
-    class MyModel(db.Model):
+    db = SQLAlchemy(model_class=EqModel)
+
+    class Language(db.Model):
         ...
 
-.. note::
+    Language.query.filter_by(name='Python') == Language.query.get(1)
```
ThiefMaster: especially for this example there is no need for a custom `__eq__` (actually, the two objects are event identical, not just equal)

docs/customizing.rst
```diff
@@ -6,75 +6,101 @@ Customizing
...
+classes to customize only those models. For example, if some models should
+track when they are created or updated::
 
-    from flask_sqlalchemy import Model, SQLAlchemy
-    from sqlalchemy import Column, DateTime
     from datetime import datetime
 
-    class TimestampedModel(Model):
-        created_at = Column(DateTime, default=datetime.utcnow)
+    class TimestampMixin:
```
ThiefMaster: `(object)` to make it new-style in python 2

flask_sqlalchemy/model.py
```diff
@@ -0,0 +1,124 @@
...
+
+        super(BindMetaMixin, cls).__init__(name, bases, d)
+
+        if bind_key is not None and hasattr(cls, '__table__'):
+            cls.__table__.info['bind_key'] = bind_key
+
+
+class DefaultMeta(
+    NameMetaMixin, BindMetaMixin, DeclarativeMeta
+):
```
ThiefMaster: this fits in one line just well, even within the 79 char limit ;)

docs/customizing.rst
```diff
@@ -121,3 +142,72 @@ by overriding the ``query_class`` class attribute on the model::
...
+        def __init__(cls, name, bases, d):
+            # custom class setup could go here
+
+            # be sure to call super
+            super(CustomMeta, cls).__init__(name, bases, d)
+
+        # custom class-only methods could go here
+
+    db = SQLAlchemy(
+        model_class=declarative_base(cls=Model, metaclass=CustomMeta)
```
ThiefMaster: I'd add `name='Model'` to avoid the model class being named `Base` in its repr

docs/customizing.rst
davidism: Good point. SQLAlchemy doesn't do anything specific for `__eq__` but the model cache results in the same instances anyway. My code I pulled this from was actually for something more complicated, need to figure out a different example.

davidism: I'll just drop the example, can't think of a good one besides the column example below.

PR #551, opened 2017-10-02
model uses existing table by name

### LINKED ISSUE #550: Model with tablename of reflected table does not have table assigned
Just updated and my full application, working previously, broke down on setting the model due to the update (https://github.com/mitsuhiko/flask-sqlalchemy/pull/541). I believe it might be a problem with this new primary key detection thing, which I'm not sure I get the point of. The previously-fine code is thus:

```
from flask import Flask

app = Flask(__name__)
db = SQLAlchemy(app)
db.reflect()

class Country(db.Model):
    __tablename__ = 'd_country'
```

Resulting in:

`InvalidRequestError: Class <class 'app.models.Country'> does not have a __table__ or __tablename__ specified and does not inherit from an existing table-mapped class.`

I'd be willing to bet that the update didn't take db.reflect() into account. I'm rolling back unless there's something else I can do, but I'd rather not have to do a declared_attr on all my tables just for this thing.

davidism:
If tables are already defined (or reflected), creating a model with an existing table name uses (or extends) that table. This worked in plain SQLAlchemy and 2.1, 2.2. `__table_cls__` needed to check for this.

fixes #550

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -2,6 +2,18 @@ Changelog
 =========
 
 
+Version 2.3.1
+-------------
+
+In development
+
+- If a model has a table name that matches an existing table in the metadata,
+  use that table. Fixes a regression where reflected tables were not picked up
+  by models. (`#551`_)
+
+.. _#551: https://github.com/mitsuhiko/flask-sqlalchemy/pull/551
+
+
 Version 2.3.0
 -------------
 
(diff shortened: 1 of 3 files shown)


davidism:
This works now:

```python
from flask import Flask

app = Flask(__name__)
app.config.update(
    SQLALCHEMY_DATABASE_URI='postgresql:///example',
)
db = SQLAlchemy(app)
db.reflect()

class User(db.Model):
    pass

print(User.query.all())
```

flipperbw:
Thanks very much for the quick turnaround,

PR #555, opened 2017-10-04
Fixes #554 representation of new model instances

### LINKED ISSUE #554: representation/printing of Models fails when model is not yet persisted to the database
When you create an new instance of a model you cannot print it or use it in logs, because the __repr__ method of Model tries to loop through the identity parameter of inspect(), which is None. 

After persisting to the database the instance can be printed.

example: 
```python
>>> from application.extensions import db
>>> from application.model.models import User
>>> u = User()
>>> print(u)
  File "<console>", line 1, in <module>
  File "/home/jeff/virtualEnvs/kongoserver/lib/python3.5/site-packages/flask_sqlalchemy/model.py", line 121, in __repr__
    pk = ', '.join(to_str(value) for value in inspect(self).identity)
TypeError: 'NoneType' object is not iterable
>>> u.id = 12345
>>> print(u)
  File "<console>", line 1, in <module>
  File "/home/jeff/virtualEnvs/kongoserver/lib/python3.5/site-packages/flask_sqlalchemy/model.py", line 121, in __repr__
    pk = ', '.join(to_str(value) for value in inspect(self).identity)
TypeError: 'NoneType' object is not iterable
>>> db.session.add(u)
>>> db.session.commit()
>>> print(u)
<User 12345>
```

Where User is defined like: 
```python
class User(db.Model):
    __tablename__ = 'application_user'

    username = db.Column(db.String(64), unique=True)
    password = db.Column(db.String(255))
```

jeff-99:
This fixes issue #554 by printing an id of "None" before a model is persited to the database

### DIFF
--- a/tests/test_model_class.py
+++ b/tests/test_model_class.py
@@ -43,7 +43,7 @@ def test_repr(db):
     db.create_all()
 
     u = User(name='test')
-    assert repr(u).startswith("<User (transient ") is True
+    assert repr(u).startswith("<User (transient ")
     db.session.add(u)
     db.session.flush()
     assert repr(u) == '<User test>'
--- a/tests/test_model_class.py
+++ b/tests/test_model_class.py
@@ -43,6 +43,7 @@ def test_repr(db):
     db.create_all()
 
     u = User(name='test')
+    assert repr(u).startswith("<User (transient ") is True
     db.session.add(u)
     db.session.flush()
     assert repr(u) == '<User test>'
--- a/flask_sqlalchemy/model.py
+++ b/flask_sqlalchemy/model.py
@@ -120,7 +120,7 @@ class Model(object):
     def __repr__(self):
         identity = inspect(self).identity
         if identity is None:
-            pk = None
+            pk = "(transient) {0}".format(id(self))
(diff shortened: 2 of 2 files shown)


davidism:
It would be nice to show something about each instance still, for example `<Model (transient) id(self)>`.

jeff-99:
Great idea, I've added the 'transient' id to the __repr__ method

davidism:
Needs a test.

Let's make it `(transient id(self))` instead so it's a little clearer the id is part of the transient mark instead of the actual key.

tests/test_model_class.py
```diff
@@ -43,6 +43,7 @@ class Report(db.Model):
     db.create_all()
 
     u = User(name='test')
+    assert repr(u).startswith("<User (transient ") is True
```
davidism: remove `is True`, it's redundant

davidism:
Looks like that latest commit was misconfigured, your git name and email don't match GitHub.

davidism:
Oops, forgot to rebase this to 2.3-maintenance. I'll cherry pick it.

PR #556, opened 2017-10-04
raise correct error when model has name but no pkey

### LINKED ISSUE #552: Simple table with no primary key raises error about missing name not key
Due to the way `__table_cls__` is deciding to create a table, we get the wrong error message when a tablename is generated for a non-inheriting model with no primary key.

```python
class User(db.Model):
    pass
```

Error is:

    sqlalchemy.exc.InvalidRequestError: Class <class '__main__.User'> does not have a __table__ or __tablename__ specified and does not inherit from an existing table-mapped class.

Error should be:

    sqlalchemy.exc.ArgumentError: Mapper Mapper|User|user could not assemble any primary key columns for mapped table 'user'

davidism:
fixes #552 

```python
class User(db.Model):
    name = db.Column(db.String)
```

```
sqlalchemy.exc.ArgumentError: Mapper Mapper|User|user could not assemble any primary key columns for mapped table 'user'
```

### DIFF
--- a/flask_sqlalchemy/model.py
+++ b/flask_sqlalchemy/model.py
@@ -73,11 +73,15 @@ class NameMetaMixin(object):
         If no primary key is found, that indicates single-table inheritance,
         so no table will be created and ``__tablename__`` will be unset.
         """
+        # check if a table with this name already exists
+        # allows reflected tables to be applied to model by name
         key = _get_table_key(args[0], kwargs.get('schema'))
 
         if key in cls.metadata.tables:
             return sa.Table(*args, **kwargs)
 
+        # if a primary key or constraint is found, create a table for
+        # joined-table inheritance
         for arg in args:
             if (
                 (isinstance(arg, sa.Column) and arg.primary_key)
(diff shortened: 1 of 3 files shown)


flask_sqlalchemy/model.py
```diff
@@ -73,18 +73,31 @@ def __table_cls__(cls, *args, **kwargs):
...
             if (
                 (isinstance(arg, sa.Column) and arg.primary_key)
                 or isinstance(arg, sa.PrimaryKeyConstraint)
             ):
                 return sa.Table(*args, **kwargs)
 
+        # if no base classes define a table, return one
+        # ensures the correct error shows up when missing a primary key
+        for base in cls.__mro__[1:-1]:
+            if '__table__' in base.__dict__:
```
ThiefMaster: I'd probably go for

```python
if not any('__table__' in b.__dict__ for b in cls.__mro__[1:-1]):
    return sa.Table(*args, **kwargs)
```

but IMO both are pretty well-readable so no strong opinon ;)

davidism: Was doing that initially for the check above this, but it looked too complicated. Kept it the same here for consistency.

PR #569, opened 2017-12-04
distinguish SQLAlchemy class from SQLAlchemy library

danidee10:
It should be `Flask SQLAlchemy` Since we're talking about the differences between `Flask SQLAlchemy` and Pure `SQLAlchemy`


ThiefMaster:
This is breaking the RSt link pointing to the class...

danidee10:
@ThiefMaster i've fixed the broken link.

davidism:
This change doesn't look correct.

PR #573, opened 2017-12-16
Retrieve all items without limit in pagination

adzhurinskij:
Hi.

Sometimes, i need retrieve all items and use big limit, for example limit=9999999. I think is not practical. This commit can do this case more easy.

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -326,7 +326,7 @@ class Pagination(object):
     def pages(self):
         """The total number of pages"""
         if self.per_page == 0:
-            pages = 0
+            pages = 1
         else:
             pages = int(ceil(self.total / float(self.per_page)))
         return pages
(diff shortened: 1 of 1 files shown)


nicolashahn:
Why are you using pagination if you just need all items? That's not what pagination is for.

ThiefMaster:
A "show all" button makes lots of sense in many applications.

adzhurinskij:
Show all - its good example.

End second example: i can use one endpoin for pagination with limits or html select without limits.

CoburnJoe:
I can see the benefits of a show all (although as @nicolashahn says, why use paging at all for this use case, unless you want API consistency with the nested data and paging metadata?).

I don't think this implementation is the way to do so. Here's an example. An application serves up an API, and uses ``flask-sqlalchemy`` to handle the paging for API consumers. By passing the query string args directly to the pagination object, it's easy for developers to integrate paging into their APIs.

With your proposed change, users of pagination in APIs would need to check their data provided to pagination, to ensure their API consumers can't bypass paging just by passing ``0`` as the ``per_page`` limit. I think a specific (optional, defaulted) flag to skip ``limit`` and ``offset`` and return all results would be far more suitable here, so your usage could look like this:

``query.paginate(page=page, per_page=per_page, all_results=True)``

Or even better:

``query.paginate(all_results=True)``

But I still think this confuses the matter - it handles pagination yet you don't want it to paginate. 🤔

davidism:
I don't think setting `per_page=0` is intuitive. Closing for reasons discussed above.

PR #583, opened 2018-01-11
Better performance on Pagination().total

Quanr:
#281


trancee:
@Quanr I think you are using the wrong `func` from SQLAlchemy. I usually import it like this:

> from sqlalchemy import func

And your code change should then look like this:

>         total = self.session.execute(
>             self.statement.with_only_columns(
>                 [func.count()]
>             ).order_by(None)).scalar()

Otherwise I think this pull request should be merged into the code base.

davidism:
Closing for the same reasons discussed in the closed linked PR.

PR #577, opened 2017-12-28
Add SQLALCHEMY_POOL_PRE_PING config

pigletfly:
Add `SQLALCHEMY_POOL_PRE_PING` config to test pool connection for liveness, which is added in SQLAlchemy 1.2.0.

Issue #120, opened 2013-03-20
changing isolation level


ssfking:
How do you change the isolation level of a session/transaction without going down to the sqlalchemy layer?

cjkeeney:
:+1:

fjbsantiago:
Did you manage to figure this out?

sh4r3m4n:
Hi! I wrote a custom SQLAlchemy class to do this, since `create_engine` is called inside Flask-SQLAlchemy code. Maybe someone could find this useful:

```python
from flask_sqlalchemy import (
    SQLAlchemy as OriginalSQLAlchemy,
    _EngineConnector
)

class SQLAlchemy(OriginalSQLAlchemy):
    """Override to fix issues when doing a rollback with sqlite driver
    See http://docs.sqlalchemy.org/en/rel_1_0/dialects/sqlite.html#serializable-isolation-savepoints-transactional-ddl
    and https://bitbucket.org/zzzeek/sqlalchemy/issues/3561/sqlite-nested-transactions-fail-with
    for furhter information"""

    # TODO: only do this on sqlite, not on postgres nor other RDBMS!

    def make_connector(self, app=None, bind=None):
        """Creates the connector for a given state and bind."""
        return CustomEngineConnector(self, self.get_app(app), bind)


class CustomEngineConnector(_EngineConnector):
    """Used by overrideb SQLAlchemy class to fix rollback issues"""

    def get_engine(self):
        # Use an existent engine and don't register events if possible
        uri = self.get_uri()
        echo = self._app.config['SQLALCHEMY_ECHO']
        if (uri, echo) == self._connected_for:
            return self._engine

        # Call original metohd and register events
        rv = super(CustomEngineConnector, self).get_engine()
        with self._lock:
            @event.listens_for(rv, "connect")
            def do_connect(dbapi_connection, connection_record):
                # disable pysqlite's emitting of the BEGIN statement entirely.
                # also stops it from emitting COMMIT before any DDL.
                dbapi_connection.isolation_level = None

            @event.listens_for(rv, "begin")
            def do_begin(conn):
                # emit our own BEGIN
                conn.execute("BEGIN")
        return rv


db = SQLAlchemy()

```

DenerKup:
Are there any progress on this? We were surprissed by postgres default of "READ COMMITED" instead of "READ REPEATED" today, but it looks there is no way to pass parameter.

brodul:
[...]

PR #597, opened 2018-02-23
Add `engine_params` parameter to `flask_sqlalchemy.SQLAlchemy` constructor

IcedNecro:
One of the most discussed topic of issues is to add some additional parameters for passing to `sqlalchemy.create_engine` function (http://docs.sqlalchemy.org/en/latest/core/engines.html#sqlalchemy.create_engine). `engine_params` dictionary is passed 'as is' to `sqlalchemy.create_engine`, so now it allows to pass such parameters as `isolation_level` (#120 ), `pool_pre_ping` (#577 )  etc.

Example of usage:

```python
from sqlalchemy.pool import StaticPool

db = SQLAlchemy(engine_params={'isolation_level': 'READ_UNCOMMITED', 'poolclass': StaticPool})
```

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -552,11 +552,17 @@ class _EngineConnector(object):
                 return self._engine
             info = make_url(uri)
             options = {'convert_unicode': True}
-            self._sa.apply_pool_defaults(self._app, options)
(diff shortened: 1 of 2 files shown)


davidism:
Duplicate of #487.

PR #600, opened 2018-03-01
add possibility to set pool_pre_ping in configuration

maximilianredt:
sqlalchemy got the option to set a `pool_pre_ping` in version 1.2 (http://docs.sqlalchemy.org/en/latest/core/engines.html). This should also be possible in flask-sqlalchemy.


pigletfly:
There is a pull  request already. #577

amahlaka:
can this be merged?
i believe that this functionality would be really useful

rsyring:
Thank you for your contribution and sorry this took so long.  We opted to go in a different direction but it will solve your use case.  Please see: https://github.com/pallets/flask-sqlalchemy/pull/684

kujiy:
Let me put an example here:

```

# > Flask-SQLAlchemy 2.4.0 
app.config["SQLALCHEMY_ENGINE_OPTIONS"] = {
    "pool_pre_ping": True,
}

# Create database connection object
db = SQLAlchemy(app)
```
https://flask-security-too.readthedocs.io/en/stable/quickstart.html

image
https://flask-sqlalchemy.palletsprojects.com/en/2.x/config/

PR #608, opened 2018-03-14
Fix exception in example on first run

MatthewScholefield:
This adds `db.create_all()` to the main block of the example code. Otherwise, database operations error on the first run because none of the tables exist.


MatthewScholefield:
Closing from lack of response.

PR #616, opened 2018-04-11
Updated #222 to work with changes done over time

Sp1tF1r3:
A lot of time has passed since #222 was published. Updated it to reflect recent changes.


harryx9x9:
Please, please add this fix to your base branch. It's crazy to have issues when using same table names in multiple databases for sharding purposes. This is how everyone I know does it ourselves included and it's one of the reasons we can't use SQLAlchemy in some applications. ++++1

PR #453, opened 2016-12-23
Improve CamelCase->snake_case function

AdamNiederer:
The current CamelCase->snake_case function is incorrect when working
with acronyms and numbers in words. This caused a pretty interesting
bug hunt today at work.

Examples (old and proposed method):

`VP9Codec` -> `v_p9_codec` (old)
`VP9Codec` -> `vp9_codec` (proposed)

`HTTP2RequestSession` -> `htt_p2_request_session` (old)
`HTTP2RequestSession` -> `http2_request_session` (proposed)

`UserST4` -> `user_s_t4` (old)
`UserST4` -> `user_st4` (proposed)

`HTTP2ClientType3EncoderParametersSSE`
-> `htt_p2_client_type3_encoder_parametersSSE` (old)
-> `http2_client_type_3_encoder_parameters_sse` (proposed)

PR #617, opened 2018-04-21
Fix camelcasing function

AdamNiederer:
See #453 - rebased on 2.3


davidism:
Thanks for working on this. We ended up using some look ahead/behind in the regex to overall simplify the function. Speed wasn't a main factor, but it seems to be just as fast as this no regex version. There's a slight difference in handling numbers, but it's ambiguous either way in some cases.

PR #625, opened 2018-06-01
add SQLALCHEMY_POOL_PRE_PING

dneuhaeuser-zalando:
This allows setting the create_engine() pool_pre_ping argument via flask configuration.

I've had some test failures locally but they appear to be unrelated as far as I can tell.


dneuhaeuser-zalando:
Oops, looks like there is already a PR for this feature: https://github.com/mitsuhiko/flask-sqlalchemy/pull/577

Issue #98, opened 2012-08-24
Document how to use plain SQLAlchemy models with the Flask-SQLAlchemy session


dmckeone:
The application I was writing required that I have models that were portable (not just for use with Flask), but I still wanted to be able to take advantage of the sessions that Flask-SQLAlchemy provided.

It would be nice if it were documented somewhere that, if you are willing to forego some of the convenience methods on the Model (like Model.query), then you can use standard SQLAlchemy models with the Flask-SQLAlchemy session.

Incidentally this approach also helps with avoiding circular imports when passing the SQLAlchemy() instance, db, around to each model.

Here is an example:

```python
# models.py

from sqlalchemy import Column, Integer, String
from sqlalchemy.ext.declarative import declarative_base

Base = declarative_base()

class User(Base):
    __tablename__ = 'users'

    id = Column(Integer, primary_key=True)
    name = Column(String(50), unique=True)
    email = Column(String(120), unique=True)

    def __init__(self, name=None, email=None):
        self.name = name
        self.email = email

    def __repr__(self):
        return '<User %r>' % (self.name)
```
```python
# app.py

from flask import Flask
from models import Base, User
from flask_sqlalchemy import SQLAlchemy

app =  Flask(__name__)
app.config['SQLALCHEMY_DATABASE_URI'] = 'sqlite:////tmp/test.db'
db = SQLAlchemy(app)

@app.before_first_request
def setup():
    # Recreate database each time for demo
    Base.metadata.drop_all(bind=db.engine)
    Base.metadata.create_all(bind=db.engine)
    db.session.add(User('Bob Jones', 'bob@gmail.com'))
    db.session.add(User('Joe Quimby', 'eat@joes.com'))
    db.session.commit()

@app.route('/')
def root():
    users = db.session.query(User).all()
    return u"<br>".join([u"{0}: {1}".format(user.name, user.email) for user in users])

if __name__ == '__main__':
    app.run('127.0.0.1', 5000) 
```

nickw444:
@mozillazg I came across this today and dirty patched it for my own usage. Not heavily tested, but would really like this to be eventually merged back into Flask-SQLAlchemy since it's useful if you're using the models in non flask-apps at the same time (or wish to modularise them). 

My patch simply adds the method `db.register_base(Base)`.

For Example
```
from models import Base
app = Flask(__name__)
db = SQLAlchemy(app)
db.register_base(Base)
```

[...]

PR #629, opened 2018-06-28
Add docs about integrating pure SQLAlchemy

tiangolo:
Add section explaining how to write pure SQLAlchemy models that can be easily imported and used in non-Flask code, and how to integrate them with Flask-SQLAlchemy for Flask code, to be able to use the Flask-specific features, like scoped sessions.

With inspiration from @dmckeone [comment](https://github.com/mitsuhiko/flask-sqlalchemy/issues/98#issue-6429516) on #98 and @gtalarico [comment](https://github.com/mitsuhiko/flask-sqlalchemy/pull/250#issuecomment-377080229) on #250

I developed and tried the code before writing the guide to make sure it "works". Also, it's actually a simplification of what I'm doing at this Flask project generator https://github.com/tiangolo/full-stack

I tried to keep it as concise as possible while still being understandable, practical and something that can be copy-pasted and tried without more modifications.

### DIFF
--- a/docs/customizing.rst
+++ b/docs/customizing.rst
@@ -188,3 +188,108 @@ generation can be disabled by defining a custom metaclass. ::
 
 This creates a base that still supports the ``__bind_key__`` feature but does
 not generate table names.
+
+
+Using pure SQLAlchemy models and combining them with Flask-SQLAlchemy
+---------------------------------------------------------------------
+
+In some cases you might want to use only the session handling from
+Flask-SQLAlchemy, but use pure SQLAlchemy for your models and querying.
+
+This would allow you to use the same models and code in Python applications
+other than Flask, but keep the Flask-specific features for Flask applications.
+
+To achieve that, create your ``Base`` class using pure SQLAlchemy::
+
+    # Inside app/db/base.py
+    # There should also be an empty file app/db/__init__.py
+
+    from sqlalchemy.ext.declarative import declarative_base, declared_attr
+
+
+    class CustomBase(object):
+        # Generate __tablename__ automatically
+        @declared_attr
+        def __tablename__(cls):
+            return cls.__name__.lower()
+
+
+    Base = declarative_base(cls=CustomBase)
+
+
+Then create your model using that ``Base``::
+
+    # Inside app/models/user.py
+    # There should also be an empty file app/models/__init__.py
+
+    from sqlalchemy import Column, Integer, String
+    from ..db.base import Base
+
+
+    class User(Base):
+        id = Column(Integer, primary_key=True, index=True)
+        name = Column(String, index=True)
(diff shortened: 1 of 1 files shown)


davidism:
Just curious, is there a reason for the emoji placeholder in the title? I've seen a bunch of PRs in the last day with that pattern. Note that they don't work in the title, they just look weird.

tiangolo:
Hehe, oops, sorry for the weird title. I didn't pay attention to it and left if like that from the commit message :grin: 

Using emoji in commit messages comes from this https://gitmoji.carloscuesta.me/ (as far as I know)

The idea is to specify what kind of action was done with a bit of cosmetic decoration.  Nothing important, really. Just an attempt to look fancy... :joy:

rsyring:
Do we have tests that mirror these examples so they know they work?

Maybe better yet, is it possible to import test content into the docs so that we could actually embed the source of tests?  That would ensure the examples we have in the docs always work.

davidism:
You can look at [Sphinx's doctest extension](http://www.sphinx-doc.org/en/master/usage/extensions/doctest.html) or [Click's doc runner](https://github.com/pallets/pallets-sphinx-themes/blob/master/src/pallets_sphinx_themes/themes/click/domain.py), but I don't think doctests are worth it right now.

tiangolo:
@rsyring just as a note/comment, although I guess it probably doesn't apply here as the docs are generated differently.

---

I'm using [MkDocs](https://www.mkdocs.org/) for [FastAPI](https://github.com/tiangolo/fastapi), I use https://github.com/cmacmackin/markdown-include to include source Python files in the docs, and most of the tests are actually run against those same Python files from the docs.

That does what you say, it ensures the sources in the docs work, and also it helps ensuring that the features are documented: if coverage is 100% and all (or most of) the tests are from docs, all (or most of) the features are documented.

The equivalent of this PR in FastAPI would be from this source file: https://github.com/tiangolo/fastapi/blob/master/docs/src/sql_databases/tutorial001.py

And the docs (that include it) are here: https://fastapi.tiangolo.com/tutorial/sql-databases/

davidism:
I was reviewing this, and it seems like it's really just saying that you can call `db.session.query()` on any model, not just ones derived from `db.Model`. I guess that's useful information, but I don't think it really warrants a section this long. This seems to be more of a mini-tutorial on creating a base class with SQLAlchemy than doing anything specific with Flask-SQLAlchemy.

You don't even need to pass `model_class=Model` the only thing that does that's relevant to the given use case is adding a `query` property, but even then if you're using this outside Flask you're probably just doing `session.query()` instead already.

Going to close this, but I'm open to adding something about being able to pass any model to `db.session.query()` if someone finds a good place for it.

PR #638, opened 2018-09-14
replace time.clock with perf_counter for py3

frostming:
Replace the timer to suppress the deprecation warning.

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -36,7 +36,10 @@ __version__ = '2.3.2'
 
 # the best timer function for the platform
 if sys.platform == 'win32':
-    _timer = time.clock
+    if sys.version_info >= (3, 3):
+        _timer = time.perf_counter
+    else:
+        _timer = time.clock
 else:
     _timer = time.time
 


flask_sqlalchemy/__init__.py
```diff
@@ -36,7 +36,10 @@
 
 # the best timer function for the platform
 if sys.platform == 'win32':
-    _timer = time.clock
+    if sys.version_info()[:2] > (3, 3):
```
chardbury: See: https://docs.python.org/3/library/time.html#time.perf_counter

```suggestion
    if sys.version_info()[:2] >= (3, 3):
```

frostming: Oh, thanks for getting into this.

MichaelF91:
I came here to suggest this, as I'm getting ugly deprecation warnings using pytest. Is there a reason why this pull request hasn't yet been accepted?

rsyring:
#662 uses `if sys.version_info >= (3, 3):` while this PR uses `sys.version_info[:2] >= (3, 3):`.  I believe both would work given tuple comparison semantics and the former is just a bit cleaner.

Also, we don't currently have CI setup and I don't have easy access to a Windows machine.  Will someone manually confirm that this code works as expected on Windows for Python 2.7 and a version of Python3 > 3.3.

frostming:
@rsyring I squashed the commits and tested on Windows PC, with Python 3.7.2 and Python 2.7.15, all pass. Thanks

rsyring:
Great, thank you.  I'll update the milestone to 2.4 as I see no reason to delay this.

rsyring:
Would you mind rebasing this on the current `2.x-maintenance` branch and change the PR to merge into that branch?  I believe that is the last step needed before I can get this merged.

Thanks.

davidism:
The command for that would be `git rebase --onto origin/2.x-maintenance origin/master deprecate-timer`, then `git push -f`, then edit the PR here to change the target.

rsyring:
@davidism ok, I'll take care of it then.  Thanks @frostming for the submission.

rsyring:
@frostming I've replaced this PR with #700 b/c I wasn't sure if I could rebase and force-push to your fork's branch.

rsyring:
@davidism I merged this PR to the maintenance branch, after rebasing again, but the PR didn't pick up that the commit was merged.  Let me know here or on Discord if you can tell what I did wrong.

Closing this PR, but it was merged.

Issue #642, opened 2018-10-05
Inconsistent performance when using factory mode and normal mode.


garenchan:
When using normal mode, we can successfully get the engine.
```
from flask import Flask
from flask_sqlalchemy import SQLAlchemy

app = Flask(__name__)
# add some configurations.
db = SQLAlchemy(app)
print(db.engine)
```
But when using factory mode, a **RuntimeError** raised.
```
from flask import Flask
from flask_sqlalchemy import SQLAlchemy

app = Flask(__name__)
# add some configurations.
db = SQLAlchemy()
db.init_app(app)
print(db.engine)
```
The traceback is as follows:
```
Traceback (most recent call last):
  File "test.py", line 17, in <module>
    print(db.engine)
  File "E:\.venv\lib\site-packages\flask_sqlalchemy\__init__.py", line 877, in engine
    return self.get_engine()
  File "E:\.venv\lib\site-packages\flask_sqlalchemy\__init__.py", line 886, in get_engine
    app = self.get_app(app)
  File "E:\.venv\lib\site-packages\flask_sqlalchemy\__init__.py", line 912, in get_app
    'No application found. Either work inside a view function or push'
RuntimeError: No application found. Either work inside a view function or push an application context. See http://flask-sqlalchemy.pocoo.org/contexts/.
```

ThiefMaster:
When using `init_app` you need to have an active Flask app context to use the extension.

PR #643, opened 2018-10-05
Fix issue #642: Record app when configuring it.

garenchan:
Fix issue #642.

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -766,6 +766,8 @@ class SQLAlchemy(object):
         of an application not initialized that way or connections will
         leak.
         """
+        self.app = app
+
         if (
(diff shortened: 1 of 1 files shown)


ThiefMaster:
This isn't done on purpose because `init_app` can be used with multiple apps.

garenchan:
@ThiefMaster， So why did SQLAlchemy records the app when I created it:
https://github.com/mitsuhiko/flask-sqlalchemy/blob/50944e77522d4aa005fc3c833b5a2042280686d3/flask_sqlalchemy/__init__.py#L675-L684

ThiefMaster:
Because in *very* small apps that do not use an app factory this provides some convenience.

garenchan:
The problem is that the two modes are inconsistent and you're not explaining them in the document.
Why don't you remove this statement from the constructor?
```
self.app = app 
```

Issue #645, opened 2018-10-10
models_committed signal triggered too early with nested transactions


michamos:
I would like to be able to determine when changes have been committed to the db so that a query from a different worker will see the new data.

It looks like I could use the `models_committed` signal, but it doesn't seem to work properly when  using [SAVEPOINT transactions](https://docs.sqlalchemy.org/en/latest/orm/session_transaction.html#using-savepoint). In that case, the signal gets sent on the `commit()` of the nested transaction, rather than the real commit of the full transaction to the database.

This can be seen on the the following example:
```python
from flask import Flask
from flask_sqlalchemy import SQLAlchemy, models_committed


app = Flask(__name__)
app.config['SQLALCHEMY_DATABASE_URI'] = 'sqlite:////tmp/test.db'
app.config['SQLALCHEMY_TRACK_MODIFICATIONS'] = True
db = SQLAlchemy(app)

class User(db.Model):
    id = db.Column(db.Integer, primary_key=True)
    username = db.Column(db.String(80), unique=True, nullable=False)
    email = db.Column(db.String(120), unique=True, nullable=False)


@models_committed.connect
def committed(*args, **kwargs):
    print('committed signal sent')


db.create_all()

with db.session.begin_nested():
    print('inside nested')
    admin = User(username='admin', email='admin@example.com')
    db.session.add(admin)
    print('added admin')
print('outside nested')
db.session.commit()
```
The behavior I would expect is that the signal gets sent when doing the `commit()` on the last line, but instead the output is:
```
inside nested
added admin
committed signal sent
outside nested
```
due to the implicit `commit()` when exiting the context manager.

One solution would be to modify https://github.com/mitsuhiko/flask-sqlalchemy/blob/50944e77522d4aa005fc3c833b5a2042280686d3/flask_sqlalchemy/__init__.py#L212 in order to detect whether the current transaction is nested, and don't do anything in that case.

Is there any other way to reliably detect when a transaction has been committed?

ticosax:
SQLAlchemy exposes [after_transaction_end](https://docs.sqlalchemy.org/en/latest/orm/events.html?highlight=after_%20commit#sqlalchemy.orm.events.SessionEvents.after_transaction_end). 
Wouldn't it be better to rely on this signal to serve your use case ?

michamos:
[...]

PR #646, opened 2018-10-15
correctly handle signals in nested transactions

michamos:
This PR fixes the issue reported in #645 by checking whether the current transaction is nested in the `before_commit` and `after_commit` handlers, and not doing anything if that's the case. Handling rollbacks of nested transaction is a bit more tricky: it requires maintaining a stack of the changes in every subtransaction, and only discarding the changes in the subtransactions that are rolled back.

I have added a test (and a fixture to work around issues in the sqlite driver) to check that the behavior (actual changes sent at the moment the full transaction is committed) is correct in two cases: when a subtransaction rollback preserves an outer transaction, and when it causes the whole transaction to be rolled back.

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -166,13 +166,14 @@ class _SessionSignalEvents(object):
     @classmethod
     def register(cls, session):
         if not hasattr(session, '_model_changes'):
-            session._model_changes = {}
+            session._model_changes = []
 
         event.listen(session, 'before_flush', cls.record_ops)
         event.listen(session, 'before_commit', cls.record_ops)
         event.listen(session, 'before_commit', cls.before_commit)
         event.listen(session, 'after_commit', cls.after_commit)
         event.listen(session, 'after_rollback', cls.after_rollback)
+        event.listen(session, 'after_transaction_create', cls.after_transaction_create)
 
     @classmethod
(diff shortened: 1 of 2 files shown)


michamos:
@davidism any chance this could be reviewed? it fixes an important bug for our use case.

michamos:
@davidism sorry to bother you again about this, but this is really an important issue for us, and we currently need to use ugly workarounds because of it (retrying tasks periodically until the real commit has happened and they get the new version when they query the DB). I am willing to work further on it and do whatever is needed to get it fixed. Could you maybe provide some insight into whether something like this (with potential modifications) could eventually be merged?

michamos:
What's the plan for this after https://github.com/pallets/flask-sqlalchemy/pull/727 ? as there is no plan to deprecate the modification tracking any more but just disable it by default, I think this fix (or something similar) should be added, as it's currently broken when using `begin_nested`.

davidism:
Duplicate of #358, an alternate implementation. Closing for the same reasons:

https://github.com/pallets-eco/flask-sqlalchemy/pull/1087 changes modification tracking, so this will no longer be mergeable. I also have the impression that how SQLAlchemy tracks nested transactions has changed, but I'm not sure. That said, after considering it more, I don't want to add more complexity to the track modifications feature. If you need other behavior, you're better off using SQLAlchemy's events directly.

PR #647, opened 2018-10-15
WIP Support testing with PostgreSQL

puntonim:


ThiefMaster:
I don't see the point here; Flask-SQLAlchemy is just a thing wrapper for SQLAlchemy, no need to test it with different databases...

puntonim:
Flask-SQLAlchemy behaves slightly differently with different db engines. This, indeed, only happens with some features like signals: see what @michamos found here: https://github.com/mitsuhiko/flask-sqlalchemy/pull/646

Also, look at the current state of this PR (https://travis-ci.org/mitsuhiko/flask-sqlalchemy/builds/441787060?utm_source=github_status&utm_medium=notification): some tests fail when tested against PostgreSQL.
Do you see the point?

puntonim:
What do you think @davidism ?

davidism:
Thanks for you're interest. At this time, I don't think the utility of these tests justifies the extra overhead in terms of test time and maintenance.

This adds a lot of overhead to run all tests in all environments with the only change being the database backend, which is already tested by SQLAlchemy. None of the current driver hacks apply to PostgreSQL.

From what I've seen of SQLAlchemy's test suite, I absolutely do not want to take on testing against multiple database backends unless absolutely necessary.

One of the failures is that PyPy can't use Psycopg2 due to C API. The other is against CPython nightly, and doesn't appear related to Flask-SQLAlchemy.

turtle321:
Ok, then I'll close this 😃

puntonim:
Ok it makes sense, then I'll close this 😄

PR #651, opened 2018-10-29
fix BindMetaMixin with polymorphic

minisotm:
AttributeError: 'NoneType' object has no attribute 'info'
flask_sqlalchemy/model.py:124: AttributeError

How to reproduce:
   
   ```
bind_key = 'polymorphic_bind_key'

        bind_key: 'sqlite:///:memory',
    }

    class Base(db.Model):
        __bind_key__ = bind_key

        __tablename__ = 'base'

        p_type = db.Column(db.String(50))

        __mapper_args__ = {
            'polymorphic_identity': 'base',
            'polymorphic_on': p_type
        }

    class Child1(Base):

        child_1_data = db.Column(db.String(50))
        __mapper_args__ = {
            'polymorphic_identity': 'child_1',
        }
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+Version 2.3.3
+-------------
+
+Fix "AttributeError: 'NoneType' object has no attribute 'info'", when using polymorphic models. (`#651`_)
+
+.. _#651: https://github.com/mitsuhiko/flask-sqlalchemy/pull/651
+
 
 Version 2.3.2
 -------------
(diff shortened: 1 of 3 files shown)


davidism:
Thanks, this looks good. It's been a while since I looked at this, was this failing because `hasattr` was returning `True` but it was because `__table__ = None`?

minisotm:
Yes, at this moment cls.__table__ is None, but attribute exists.

davidism:
Would you add an entry in the changelog? Then I can merge.

minisotm:
done

minisotm:
Oops, version 2.3.2 or 2.3.3?

davidism:
Hmm, didn't realize there wasn't a changelog already. Just create a new section with the next version number and no release date.

dirkesquire:
I ran into exactly the same problem, so I need this fix.
I see you merged in a fix in October, but when will you be releasing it onto pypi.org?
Currently pypi.org is still showing 2.3.2.

PR #665, opened 2019-01-21
Move assigment with app in init_app_function

heckad:
I found a bug when i wrote 
```
app = Flask(__name__)
db.init_app(app)

User.query.all()
```
errore was 
```
RuntimeError: No application found. Either work inside a view function or push an application context. See http://flask-sqlalchemy.pocoo.org/contexts/.
```
When I added `db.app = app` above `db.init_app(app)`  I solved the problem.

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -680,7 +680,6 @@ class SQLAlchemy(object):
         self.session = self.create_scoped_session(session_options)
         self.Model = self.make_declarative_base(model_class, metadata)
         self._engine_lock = Lock()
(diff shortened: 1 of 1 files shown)


ThiefMaster:
Not a bug, please check the existing issues why this is intended behavior.

heckad:
Wich issue?

ThiefMaster:
https://github.com/mitsuhiko/flask-sqlalchemy/search?q=no+application+found&type=Issues

also mentioned in the docs: http://flask-sqlalchemy.pocoo.org/2.3/contexts/

PR #679, opened 2019-03-05
Test with different versions of sqlalchemy and address a deprecation warning

### LINKED ISSUE #671: Deprecation Warning with SQLAlchemy 1.3 beta
This is probably a bit premature, but when using SQLAlchemy 1.3.0b2 I encountered a deprecation warning at https://github.com/pallets/flask-sqlalchemy/blob/master/flask_sqlalchemy/__init__.py#L157.

This can be easily addressed by replacing
```
info = getattr(mapper.mapped_table, 'info', {})
```
with:
```
info = getattr(mapper.persist_selectable, 'info', {})
```

Also, what are the plans for updating for 1.3 compatibility in general?

ticosax:
fixes #671

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,9 @@ language: python
 matrix:
   include:
     - python: 3.6
-      env: TOXENV=py,codecov
+      env: TOXENV=py-sqla1.2,codecov
+    - python: 3.6
+      env: TOXENV=py-sqla1.3,codecov
     - python: 3.5
       env: TOXENV=py,codecov
     - python: 3.4
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -154,7 +154,12 @@ class SignallingSession(SessionBase):
     def get_bind(self, mapper=None, clause=None):
         # mapper is None if someone tries to just get a connection
         if mapper is not None:
-            info = getattr(mapper.mapped_table, 'info', {})
+            try:
+                # sqlalchemy >= 1.3
+                persist_selectable = mapper.persist_selectable
+            except AttributeError:
+                persist_selectable = mapper.mapped_table
+            info = getattr(persist_selectable, 'info', {})
             bind_key = info.get('bind_key')
             if bind_key is not None:
                 state = get_state(self.app)
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -560,7 +560,7 @@ class _EngineConnector(object):
             if (uri, echo) == self._connected_for:
                 return self._engine
             info = make_url(uri)
-            options = {'convert_unicode': True}
+            options = {}
             self._sa.apply_pool_defaults(self._app, options)
             self._sa.apply_driver_hacks(self._app, info, options)
             if echo:
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,7 @@
 [tox]
 envlist =
-    py{36,35,34,27,py}
+    py36-sqla{1.2,1.3}
+    py{35,34,27,py}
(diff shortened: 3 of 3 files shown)


ticosax:
Draft PRs don't trigger build apparently, so let's trigger a build.

allardhoeve:
Tested on Python 3.7 with sqlalchemy 1.3.0. 

Works for me! Please merge. 👍

davidism:
You're adding way too many test envs here.

davidism:
You're doubling the number of envs to cover one changed line.

ticosax:
for travis do you mean ?
If I don't do it only one version of sqlalchemy gets tested
https://travis-ci.org/pallets/flask-sqlalchemy/jobs/502480830

davidism:
You need one extra env here, not 5.

ticosax:
ok I'll see what I can do, thanks.

ticosax:
sorry @davidism I don't know how to make the travis matrix simpler

Could you point me to some example maybe ?

```
tox -l
py36-sqla1.2
py36-sqla1.3
py35-sqla1.2
py35-sqla1.3
py34-sqla1.2
py34-sqla1.3
py27-sqla1.2
py27-sqla1.3
pypy-sqla1.2
pypy-sqla1.3
py36-lowest
py34-lowest
py27-lowest
pypy-lowest
docs_html
coverage_report
```

davidism:
Just don't add most of the things you added?

ticosax:
you need to be more specific, you don't want I touch the travis.yaml file ? the tox.ini file ? both ?

ticosax:
Side note: that's not a sane way of communication to edit you previous comments, because we don't get notified about the change and we appear to be dumb of asking an obvious question, which was not the case at the time it was written.

So, let's play the guessing game, because there is many ways to interpret your directions. Would you like to see in the travis matrix only those two combinations added ? 
```
py36-sqla1.2
py36-sqla1.3
```
And for the rest of python versions we don't care which version of sqlalchemy is picked ?

Please confirm.
Thanks

jpassaro:
seems to me the question is, since the change calls for testing against multiple versions of SQLA, what combination of python versions and SQLA versions gets you the appropriate coverage without making CI too heavy? Do you want to test all python versions against SQLA<=1.3, and only 3.7 against SQLA>1.3? All python versions against the newer SQLA and only 3.7 against SQLA<=1.3? Or should it be only python versions supported by the specific SQLA version? (I suspect that would still be a much larger matrix.) Regardless it seems like a decision for the maintainer.

ticosax:
@jpassaro totally makes sense, I'm up for it, I just wish to receive clear directions, because obviously I'm very bad at guessing.

.travis.yml
```diff
@@ -4,15 +4,25 @@ language: python
 matrix:
   include:
     - python: 3.6
-      env: TOXENV=py,codecov
+      env: TOXENV=py36-sqla1.2,codecov
```
barrachri: @ticosax you can keep the same number of python versions as before and substitute the previous `env` with `TOXENV=py-sqla{1.2,1.3}, codecov `

ticosax: thanks! that's what I was looking for when I first understood the feedback.
But today I'm not sure that's what the maintainers are after anymore. I'll probably go into a direction on testing sqalchemy 1.2 and 1.3 only for python3.6 instead of all supported python versions, as it seems preferred by the maintainers.

ticosax:
@davidism please can you have another look ?

rsyring:
Thanks for getting us started on this.  

https://github.com/pallets/flask-sqlalchemy/pull/687 will address testing different SA versions
There is a commit in https://github.com/pallets/flask-sqlalchemy/pull/684 that will address the deprecation warning.

Issue #681, opened 2019-03-07
convert_unicode deprecation warning with SQLAlchemy 1.3


zgoda:
> _venv/lib/python3.7/site-packages/sqlalchemy/dialects/sqlite/base.py:1433: SADeprecationWarning: The create_engine.convert_unicode parameter and corresponding dialect-level parameters are deprecated, and will be removed in a future release.  Modern DBAPIs support Python Unicode natively and this parameter is unnecessary.
>     default.DefaultDialect.__init__(self, **kwargs)
> 

I guess this comes from `get_engine()` (https://github.com/pallets/flask-sqlalchemy/blob/master/flask_sqlalchemy/__init__.py#L558).

zgoda:
More on the context https://github.com/zzzeek/sqlalchemy/commit/4c2c2c40fde17c85013e00a6f3303a99e2b32c12#diff-d84f03db831ca7bb23f4dd0ea1ef3e6d

davidism:
Duplicate of #671

zgoda:
Lol.

PR #684, opened 2019-03-08
make SA engine configuration more flexible

### LINKED ISSUE #166: make it possible to pass additional options to create_engine

### LINKED ISSUE #671: Deprecation Warning with SQLAlchemy 1.3 beta

rsyring:
fixes #166 - make it possible to pass additional options to create_engine()
fixes #671, #681 - deprecation warnings from SA 1.3

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -553,18 +553,34 @@ class _EngineConnector(object):
             echo = self._app.config['SQLALCHEMY_ECHO']
             if (uri, echo) == self._connected_for:
                 return self._engine
-            info = make_url(uri)
-            options = {'convert_unicode': True}
-            self._sa.apply_pool_defaults(self._app, options)
-            self._sa.apply_driver_hacks(self._app, info, options)
-            if echo:
-                options['echo'] = echo
-            self._engine = rv = sqlalchemy.create_engine(info, **options)
+
+            sa_url = make_url(uri)
+            options = self.get_options(sa_url, echo)
+            self._engine = rv = self._sa.create_engine(sa_url, options)
+
             if _record_queries(self._app):
                 _EngineDebuggingSignalEvents(self._engine,
                                              self._app.import_name).register()
+
             self._connected_for = (uri, echo)
+
             return rv
+
+    def get_options(self, sa_url, echo):
+        options = {'convert_unicode': True}
+        self._sa.apply_pool_defaults(self._app, options)
+        self._sa.apply_driver_hacks(self._app, sa_url, options)
+        if echo:
+            options['echo'] = echo
+
+        # Give the config options set by a developer explicitly priority
+        # over decisions FSA makes.
+        options.update(self._app.config['SQLALCHEMY_ENGINE_OPTIONS'])
+
+        # Give options set in SQLAlchemy.__init__() ultimate priority
+        options.update(self._sa._engine_options)
+
+        return options
 
 
 def get_state(app):
@@ -644,6 +660,12 @@ class SQLAlchemy(object):
     to be passed to the session constructor.  See :class:`~sqlalchemy.orm.session.Session`
     for the standard options.
 
+    The ``engine_options`` parameter, if provided, is a dict of parameters
+    to be passed to create engine.  See :func:`~sqlalchemy.create_engine`
+    for the standard options.  The values given here will be merged with and
+    override anything set in the ``'SQLALCHEMY_ENGINE_OPTIONS'`` config
+    variable or othewise set by this library.
+
     .. versionadded:: 0.10
        The `session_options` parameter was added.
 
@@ -663,6 +685,9 @@ class SQLAlchemy(object):
 
     .. versionchanged:: 2.1
        Utilise the same query class across `session`, `Model.query` and `Query`.
+
+    .. versionadded:: 2.4
+       The `engine_options` parameter was added.
     """
 
     #: Default query class used by :attr:`Model.query` and other queries.
@@ -671,7 +696,8 @@ class SQLAlchemy(object):
     Query = None
 
     def __init__(self, app=None, use_native_unicode=True, session_options=None,
-                 metadata=None, query_class=BaseQuery, model_class=Model):
+                 metadata=None, query_class=BaseQuery, model_class=Model,
+                 engine_options=None):
 
         self.use_native_unicode = use_native_unicode
         self.Query = query_class
@@ -679,6 +705,7 @@ class SQLAlchemy(object):
         self.Model = self.make_declarative_base(model_class, metadata)
         self._engine_lock = Lock()
         self.app = app
+        self._engine_options = engine_options or {}
         _include_sqlalchemy(self, query_class)
 
         if app is not None:
@@ -790,6 +817,7 @@ class SQLAlchemy(object):
         track_modifications = app.config.setdefault(
             'SQLALCHEMY_TRACK_MODIFICATIONS', None
         )
+        app.config.setdefault('SQLALCHEMY_ENGINE_OPTIONS', {})
 
         if track_modifications is None:
             warnings.warn(FSADeprecationWarning(
@@ -819,7 +847,7 @@ class SQLAlchemy(object):
         _setdefault('pool_recycle', 'SQLALCHEMY_POOL_RECYCLE')
         _setdefault('max_overflow', 'SQLALCHEMY_MAX_OVERFLOW')
 
-    def apply_driver_hacks(self, app, info, options):
+    def apply_driver_hacks(self, app, sa_url, options):
         """This method is called before engine creation and used to inject
         driver specific hacks into the options.  The `options` parameter is
         a dictionary of keyword arguments that will then be used to call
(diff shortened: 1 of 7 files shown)


rsyring:
Current Travis failures:

- Python 3.3 fails due to tox supporting 3.4+
- Python 2.6 fails due to SQLAlchemy needing Py 2.7+
- docs_html: /home/travis/build/pallets/flask-sqlalchemy/.tox/docs_html/lib/python3.6/site-packages/flask_sqlalchemy/__init__.py:docstring of flask_sqlalchemy.SignallingSession.get_bind:63:Unknown 
interpreted text role "paramref".

@davidism I'm not sure where FSA sits in terms of Python version support.  Created an issue to discuss that: https://github.com/pallets/flask-sqlalchemy/issues/685

The docs error I can look into but I figured I'd see if you knew what the problem is.  This PR doesn't change that file, so I think the problem existed previously.

rsyring:
When https://github.com/pallets/flask-sqlalchemy/pull/687 is approved and merged, I'll rebase this PR on that branch and that should resolved all the CI issues.

rsyring:
I need to review how/why `get_engine()` caches the engine before this is merged.

rsyring:
> I need to review how/why get_engine() caches the engine before this is merged.

I'd like to punt on that discussion and have opened https://github.com/pallets/flask-sqlalchemy/issues/698 for discussion.

@davidism I've rebased this PR, Travis is passing, and I believe the commits in this PR are distinct enough to not squash.  Let me know if you would like me to change anything and/or if I'm good to merge.

tests/test_config.py
```diff
@@ -0,0 +1,163 @@
+import mock
```
davidism: Use pytest's `monkeypatch` fixture rather than adding a dependency on mock.

flask_sqlalchemy/utils.py
```diff
@@ -0,0 +1,20 @@
+
+from pkg_resources import parse_version
```
davidism: Don't use `pkg_resources` here, it forces a dependency on setuptools. Use [`packaging.version.parse`](https://packaging.pypa.io/en/latest/version/#packaging.version.parse) if you really need to, but I'd rather not add a new dependency. For what's needed here, you can probably get away with `tuple(int(x) for x in sqlalchemy.__version__.split(".")[:2]) < (1, 3)`.

flask_sqlalchemy/__init__.py
```diff
@@ -553,19 +561,38 @@ def get_engine(self):
...
                 _EngineDebuggingSignalEvents(self._engine,
                                              self._app.import_name).register()
+
             self._connected_for = (uri, echo)
+
             return rv
 
+    def get_options(self, sa_url, echo):
+        options = {}
+        if utils.sqlalchemy_version('<', '1.3'):
```
davidism: Reading the deprecation information on this, it seems acceptable to just remove the option altogether, rather than trying to examine SQLAlchemy's version. As their docs say, the various DB-APIs have supported Unicode on 2 and 3 for a while now.

tests/test_config.py
ThiefMaster: I don't think monkeypatch works when using the advanced features of mock (inspecting calls to the mocked function etc) unless you want to reimplement this logic in the test (which is ugly IMO). But I'd use pytest-mock instead of the decorators.

davidism: It looked like monkeypatch would be fine for the uses here, I agree (pytest-)mock is more appropriate for more complex tasks.

fuhrysteve: can also get fancy and only include mock for py2.

```python
try:
    from unittest import mock
except ImportError:
    # py2
    import mock
```

(I also find mock much more useful & less confusing than pytest's monkeypatch in almost all cases)

flask_sqlalchemy/__init__.py
rsyring: Done locally.

flask_sqlalchemy/utils.py
rsyring: Don't need this anymore since the above change request removed the only usage of needing to know the SA version.

rsyring: oops, nm, need that in the tests.

rsyring: k, resolved locally with no dependency.

tests/test_config.py
rsyring: If the only argument is to not use mock b/c it adds a dependency, `TestCreateEngine` below uses mock.patch extensively to inspect call args.  I've looked at the docs for [pytest monkeypatch](https://docs.pytest.org/en/latest/reference.html#_pytest.monkeypatch.MonkeyPatch) and don't believe it's going to support that use case.

So, if we end up using mock for those tests, might as well use it here and keep things consistent.  Also, I agree with @fuhrysteve that the mock API feels a lot cleaner and obvious than what pytest's offering.

I personally don't think we should be afraid of adding dependencies when they provide a benefit and IMO, this is one of those cases.  I also use mock extensively in my day job and will be most productive with it in future FSA updates.

rsyring:
Code updated to address two review requests.  The lingering one with respect to mock usage remains.  If the status quo is acceptable, then this is ready for merge and I'll prep a 2.4 release for @davidism's review.

rsyring:
The doc tests are failing in CI with:

```
ERROR: InvocationError for command '/home/travis/build/pallets/flask-sqlalchemy/.tox/docs/bin/sphinx-build -W -b linkcheck -d /home/travis/build/pallets/flask-sqlalchemy/.tox/docs/tmp/doctrees docs docs/_build/linkcheck' (exited with code 2)
```

That tox run passes when I test it locally, maybe GitHub is having some issues.  Anyway, I don't think that failure should hold up the merge of this PR.  I'll kick off the job a couple more times to see if I can get it to pass.

davidism:
Are there any separate features we can deprecate and roll into this? If so, I'd like those deprecation warnings to be in the next release.

aydumoulin:
@davidism The three deprecation that arose from fast forwarding `sqlalchemy` to `1.3+` in my postgres flavor stack were:
- `SADeprecationWarning`: The create_engine.convert_unicode parameter and corresponding dialect-level parameters are deprecated, and will be removed in a future release.  Modern DBAPIs support Python Unicode natively and this parameter is unnecessary. `default.DefaultDialect.__init__(self, **kwargs)`
- `SADeprecationWarning`: Use .persist_selectable, `info = getattr(mapper.mapped_table, 'info', {})`
- `SADeprecationWarning`: The 'postgres' dialect name has been renamed to 'postgresql'. `loader = self.auto_fn(name)`

davidism:
That's not what I'm referring to, those are deprecations in SQLAlchemy. I'm referring to features that we want to deprecate.

woodb:
With the recent publishing of security vulnerabilities in SQLAlchemy `v1.2.17` through `v1.3.0b2` ([CVE-2019-7164](https://nvd.nist.gov/vuln/detail/CVE-2019-7164)) getting pushed out to folks via GitHub alerts, this issue is becoming a bigger deal for users that want to upgrade to `v1.3.x`.

With these security issues abound, I would like to suggest that we consider expediting this merged and released (i.e. without [waiting for more stuff to put into the next release](https://github.com/pallets/flask-sqlalchemy/pull/684#issuecomment-479546813))

I haven't read the code here, so not sure if a patch version is appropriate, but really would like to see this fix get out so that my test suite can have useful output again, hah.

If there's anything that still needs to be done on this PR to get #671 and #681 closed, please let us know. Thanks!

rsyring:
From @davidism regarding his comment on deprecations above: 

> was specifically talking about engine settings in that comment
> Did any individual config options get added before that could be rolled up? I think ECHO is the only one to keep since it's good for debugging. http://flask-sqlalchemy.pocoo.org/2.3/config/
> There are mostly pool-related config keys.
> And NATVIE_UNICODE was no longer needed, right?
> So we might want to issue deprecation warnings if they're set, and add their value to ENGINE_OPTIONS for now.

I'll be doing that work shortly and then prepping a release.

fuhrysteve:
@woodb just to be clear, there are no known security vulnerabilities related to [CVE-2019-7164](https://nvd.nist.gov/vuln/detail/CVE-2019-7164) in any currently supported major versions of sqlalchemy.

swuecho:
Thanks! especially for SQLALCHEMY_ENGINE_OPTIONS.

billyfung:
ditto^
engine options works great for psycopg2 use_batch_mode=True

PR #693, opened 2019-03-11
fix DeprecationWarning: time.clock

Inconnu08:
fix `DeprecationWarning: time.clock has been deprecated in Python 3.3 and will be removed from Python 3.8: use time.perf_counter or time.process_time instead`

`time.clock` has been replaced with `time.perf_counter` in the code.


davidism:
Duplicate of #638

PR #695, opened 2019-03-13
Fantix/db.transaction block

ticosax:
Rebased,  squashed and pytest-ified version of #447 
Thanks @fantix for this contribution


docs/config.rst
```diff
@@ -63,6 +63,19 @@ A list of configuration keys currently understood by the extension:
...
+                                   entering `with db.transaction()` with a
+                                   session with `autocommit=False` will cause
+                                   a rollback first. Oppositely ``False`` will
+                                   stop the entering rollback from happening.
+``SQLALCHEMY_NESTED_TRANSACTION``  Controls the default implementation of
+                                   nested transactions, ``False`` to use
+                                   programed subtransactions (default, the
+                                   same as ``begin(subtransactions=True)`` in
+                                   SQLAlchemy), ``True`` to use DB savepoint
+                                   (the same as ``begin_nested``).
```
davidism: I'm not sure either of these make sense as config. Is it something you'd expect to change between different deployments of the same application? It seems to make more sense as args to the extension, since it probably affects how you'd code your queries. @fantix

fantix: Oh yes, args to the extension work just fine for me - it's been quite a while, but I think I didn't find a better place to put that switch and simply left it in the config, not really a decision affected by the practice. Please feel free to change it to whatever fits better, I do appreciate that ❤️

ticosax: It seems there is consensus to remove it. So I'm going to.

docs/config.rst
```diff
@@ -63,6 +63,13 @@ A list of configuration keys currently understood by the extension:
                                    that it will be disabled by default in
                                    the future.  This requires extra memory
                                    and should be disabled if not needed.
+``SQLALCHEMY_ISOLATE_TRANSACTION`` Controls the default behavior of root
```
davidism: Why did you remove only one of the options?

ticosax: Because I understood it like this.
I can remove it as well.

ticosax: done

davidism:
Both SQLAlchemy and this extension have changed significantly since this was created. I am not convinced that Flask-SQLAlchemy needs to do anything special here, SQLAlchemy's transaction support should work. Also, this autocommit behavior seems related to the commit-on-teardown feature that is being removed.

PR #687, opened 2019-03-08
Update docs and tox

rsyring:
- refs #685:
  - dropping support in CI for older Python versions
  - adding CI support for Python3.7
  - refactoring tox/CI as needed
- fixes #686 - tox docs build failing
- fixes #683 - fix references to version 3.0 in SQLAlchemy docstring
- improves readme and documentation including better checks in CI
- Allow to pass args to pytest through tox - replaces #696

PR #696, opened 2019-03-13
Allow to pass args to pytest through tox

ticosax:
example:
```
tox -e py35 -- --sw
```


rsyring:
I'm assuming the no argument version works as before?

```tox -e py35```

ticosax:
> I'm assuming the no argument version works as before?

Yes 100%

rsyring:
Thanks for the contribution.  This will be available shortly through #687.

PR #697, opened 2019-03-14
Remove convert_unicode option, as it is being depreciated

Drvanon:
Hi,
I am getting the following error continually:

`/home/robin/.local/share/virtualenvs/gridt-server-ZYPDaXQg/lib/python3.7/site-packages/sqlalchemy/dialects/sqlite/base.py:1433: SADeprecationWarning: The create_e
ngine.convert_unicode parameter and corresponding dialect-level parameters are deprecated, and will be removed in a future release.  Modern DBAPIs support Python
Unicode natively and this parameter is unnecessary.
`
This should be the cause of that.


davidism:
Handled by #684

PR #700, opened 2019-03-14
Windows: replace time.clock with perf_counter for py3

rsyring:
original author: frostming
replaces #638


rsyring:
We are keeping #638.

PR #715, opened 2019-04-18
Prepare 2.4 release

rsyring:
@davidism this is my attempt to prepare the project for a release.  I used [pyp](https://github.com/level12/pyp/) to automatically bump the version and generate the changelog.  As a result, the changelog is in a slightly different format than the previous versions.  

Considering the changelog was automatically generated and slightly tweaked in a just a couple minutes, I can personally live with the change in formatting to get a less time consuming release process.  I can revisit if you disagree.

I'm not sure if the `make-release.py` script is still used and, if it is, then it looks like I'll need to discard the changes regarding version or change that release script if you like what pyp does in terms of auto-generating the changelog.


davidism:
`make-release` is no longer used, but most of the patterns still are. Let's stick with the same style used in other Pallets projects, for all these changes.

The changelog should link to the issue or pr, not the commit. I've started using sphinx-issues for that. I will start looking into towncrier for better managed changelogs at PyCon next month.

davidism:
I need to do some final cleanup to get the metadata consistent with other projects, as well as the docs using pallets-sphinx-themes and building on RTD.

ThiefMaster:
strong -1 on generating the changelog from the commits. some commits might not belong there (typos, refactoring that doesn't change anything visible on the outside, CI/test stuff, etc).

when reading a changelog i like to quickly see relevant things: breaking changes, improved/new features and bugfixes

so I'd prefer sticking with human-written changelog entries

rsyring:
> strong -1 on generating the changelog from the commits.

FWIW, there are two things that have made this helpful in my projects even though I agree there is a potential for the changelog to be messy:

1. We usually use a PR for everything, cleanup the commit history in the PR as part of our review, and use a merge commit to merge (not fast forward).  The commit history therefore is often very close to a history of merged PRs.
2. The changelog can always be edited after it is generated.  Due to the above point, I have found it less time consuming to generate the changelog and tweak for readability than to generate the changelog completely by hand.

davidism:
changelog cherry picked to #717

Issue #688, opened 2019-03-08
Add flake8 and/or other style helpers


rsyring:
See https://github.com/pallets/werkzeug/blob/master/.pre-commit-config.yaml for reference.

davidism:
This requires reducing the pr backlog to near zero then applying it in one shot. Otherwise everything had merge conflicts.

davidism:
#827

PR #716, opened 2019-04-18
Add flake8

rsyring:
refs #688 

This is just to show how little work was needed in the code to get flake8 passing.  If @davidism approves, I could add flake8 to CI and merge this.  I don't think it's impact on future PR requests will be that big.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+insert_final_newline = true
+trim_trailing_whitespace = true
+end_of_line = lf
+charset = utf-8
+max_line_length = 88
+
+[*.{yml,yaml,json,js,css,html}]
+indent_size = 2
--- a/.flake8
+++ b/.flake8
@@ -1,7 +1,0 @@
-[flake8]
-max-line-length = 100
-exclude=.git,.tox,dist,doc,*egg,build
-# E128 - Continuation line under-indented for visual indent.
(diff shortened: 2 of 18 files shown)


davidism:
Let's use Werkzeug's config, I'd like all projects to be consistent. Line length will be 88 to match Black's formatting. We use the flake8-bugbear extension as well for some checks.

davidism:
The problem with adding CI for this is that any time you merge in an existing PR that doesn't conform, you break CI for all new PRs. The existing PRs don't have the CI config, so you don't see an error until the code you merged shows up in a PR that does have the CI config.

rsyring:
> The problem with adding CI for this now is that any time you merge in an existing PR that doesn't conform, you break CI for all new PRs

I usually rebase PRs onto the existing branch before merging.

> This is why I generally like to wait until there are no open PRs

Given the number of open PRs even after I cleaned up what I could, and the small amount of time available to maintain this package, I'd rather not wait until all the open PRs are merged and just focus on fixing the PRs prior to merge.

> Or we can fully commit to the Pallets style now and apply the other pre-commit hooks like Black as well, with the understanding that we'll need to clean up existing PRs.

I've never needed anything more than flake8, so can't speak to the other tools.  If that means we get a style checker soon, then I'm for it.  :)

davidism:
Going to close this since I'll be doing this along with a bunch of other style stuff when I add the pre-commit config from the other projects and drop Python 2.

PR #720, opened 2019-04-19
adapt flaskr example to sqlalchemy

davidism:
This takes the Flaskr example used in the Flask tutorial, and adapts it from plain SQL to SQLAlchemy. At some point someone could write a short tutorial that explains what's different.


examples/flaskr/flaskr/auth.py
```diff
@@ -0,0 +1,121 @@
...
+from flask import session
+from flask import url_for
+from sqlalchemy.ext.hybrid import hybrid_property
+from werkzeug.security import check_password_hash
+from werkzeug.security import generate_password_hash
+
+from flaskr import db
+
+
+class User(db.Model):
```
ThiefMaster: do we really want to mix models and blueprints in the same file?

davidism: You mean make the blueprint a package with `views` and `models` instead?

PR #721, opened 2019-04-24
check paginate  args page, per_page type

akhilharihar:
Documentation for flask_sqlalchemy paginate class

When error_out is True (default), the following rules will cause a 404 response:
- No items are found and page is not 1.
- page is less than 1, or per_page is negative.
- page or per_page are not ints.

But the paginate class does not have an implementation to check if the given page and per_page arguments are int.

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -477,6 +477,22 @@ class BaseQuery(orm.Query):
         if max_per_page is not None:
             per_page = min(per_page, max_per_page)
 
+        try:
+            page = int(page)
+        except (TypeError, ValueError):
+            if error_out:
+                abort(404)
+
+            page = 1
+
+        try:
+            per_page = int(per_page)
+        except (TypeError, ValueError):
+            if error_out:
+                abort(404)
+
+            per_page = 20
+
         if page < 1:
             if error_out:
                 abort(404)


rsyring:
@akhilharihar Thank you for the contribution.  Explicit errors are always better than explicit ones.  After an initial review of the code, I'm a +1 on getting this merged.  However, will you please add tests for this PR.  If you need suggestions on what to test, let me know.

Also, could you detail how this PR affects backwards compatibility?  Are we throwing errors where no errors would have show up previously or are we just throwing better errors?  This could affect whether this PR should target `master` or the `2.x` branch.

akhilharihar:
Thank for the consideration. 

If the page and per_page arguments in paginate methods are not validated or if they are passed directly from a request object, we would end up with `TypeError: '<' not supported between instances of 'str' and 'int'`. 

This commit removes the need for a developer to validate the page, per_page params and return response based on error_out argument, making in backwards compatible.

In this case, we are throwing 400 error if error_out is true or change page, per_page values to 


```
 

So instead of letting the developer vali

takwas:
> @akhilharihar Thank you for the contribution. Explicit errors are always better than explicit ones. After an initial review of the code, I'm a +1 on getting this merged. However, will you please add tests for this PR. If you need suggestions on what to test, let me know.
> 
> Also, could you detail how this PR affects backwards compatibility? Are we throwing errors where no errors would have show up previously or are we just throwing better errors? This could affect whether this PR should target `master` or the `2.x` branch.

@rsyring: I think you meant "Explicit errors are always better than **implicit** ones".

akhilharihar:
Looks like the CI build failed for this PR. I did check the details, unfortunately, the error is being triggered at another location and not at the lines the changes are made. Can you run the build on this PR again?

davidism:
fixed in #1087

PR #740, opened 2019-05-21
Use unique MetaData object for each bind #26 #222

delijati:
Updated to current flask-sqlalchemy


davidism:
This appears to have lost all author information from #222, making it look like you're the original author. It can be helpful to refresh a PR if the code changes significantly, but that doesn't seem to be the case here, and there's no commit that shows the diff between the old and new code.

delijati:
files where moved; diff imposible; old pull request is locked; add inform authors

PR #742, opened 2019-05-23
[WIP] Allow Flask-SQLAlchemy to work with external transactions

### LINKED ISSUE #666: Multiple binds doesn't connect to external transaction in tests
I have a Flask app configured only with multiple binds using `SQLALCHEMY_BINDS` config option. The app runs fine, however,  I am struggling to get external transaction to rollback in tests.

I have following setup:

```python
# Config
SQLALCHEMY_BINDS = {
    'db1': 'sqlite:////my.db'
}

# My model
class MyModel(db.model):
    __bind_key__ = 'db1'
    email = db.Column(db.String)

# Fixtures
@pytest.fixture(scope='session')
def app():
    """Session wide Flask app"""
    return create_app()


@pytest.fixture(scope='session')
def db(app, request):
    """Session-wide test database."""
    _db.app = app
    _db.create_all()

    yield _db

    _db.drop_all()

    # On windows close the file first.
    os.close(app.config['TEST_DB_FD'])
    os.unlink(app.config['TEST_DB'])

@pytest.fixture(scope='function')
def session(db, request):
    """Creates a new database session for a test."""
    connection = db.engine.connect()
    transaction = connection.begin()

    options = dict(bind=connection, binds={})

    _session = db.create_scoped_session(options=options)

    db.session = _session

    yield _session

    _session.remove()

    transaction.rollback()
    connection.close()

@pytest.fixture
def data(session):
    session.add(MyModel(email='bob'))
    session.add(MyModel(email='alice'))

    session.comit()
```

After test is done, the models `bob` and `alice` are still in the database.

As discussed in #468 it appears there is an issue with joining to external transaction, very useful in tests. I tried the "hack":
```python
class _dict(dict):
    def __nonzero__(self):
        return True
```

But still nothing. I also tried using `binds={}` and no binds and the behaviour is different but still rollback doesn't work. I ended up going through with the debugger and it looks like external transaction is not detected, so it creates a new one for my fixture and commits it to disk, instead of external transaction.

* sqlalhemy==1.2.14
* flask-sqlalchemy==2.3.2

kevin-brown:
The goal of this is to fix #666.

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -146,26 +146,6 @@ class SignallingSession(SessionBase):
             bind=bind, binds=binds, **options
         )
 
-    def get_bind(self, mapper=None, clause=None):
-        """Return the engine or connection for a given model or
-        table, using the ``__bind_key__`` if it is set.
-        """
-        # mapper is None if someone tries to just get a connection
-        if mapper is not None:
-            try:
-                # SA >= 1.3
-                persist_selectable = mapper.persist_selectable
-            except AttributeError:
-                # SA < 1.3
-                persist_selectable = mapper.mapped_table
-
-            info = getattr(persist_selectable, 'info', {})
-            bind_key = info.get('bind_key')
-            if bind_key is not None:
-                state = get_state(self.app)
-                return state.db.get_engine(self.app, bind=bind_key)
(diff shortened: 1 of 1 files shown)


davidism:
Should be fixed by #1087, or at least significantly changed.

PR #580, opened 2018-01-09
Pagination dict() support

nicolashahn:
For issue https://github.com/mitsuhiko/flask-sqlalchemy/issues/482 

Credit goes to @protream, I just implemented

Now you can do this: 
```
>>> p = Pagination('a',1,20,100,[])
>>> dict(p)
{'page': 1, 'pages': 5, 'total': 100, 'has_prev': False, 'next_num': 2, 'has_next': True, 'items': []}
>>> p['page']
1
```

PR #748, opened 2019-05-31
Issue #482: Add as_dict method to Pagination object and associated tests

rtilk89:
This is a stab at addressing #482. 

The problem with #485 seemed to be the serialization of the items (which from my understanding is a collection of rowProxy's from sqlalchemy). I attempted to add basic support for serialization of basic items people would query from the DB: int, float, datetimes (iso 8601 format), strings. 

The problem with #580 seemed to be lack of tests.

Look forward to hearing your feedback :)

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -374,40 +374,39 @@ class Pagination(object):
         return self.page + 1
 
     @staticmethod
-    def _serialize_item(value):
-        """Handles specific JSON serialization of different datatypes"""
-        if value is None:
-            return 'null'
-        if isinstance(value, bool):
-            return str(value).lower()
-        if isinstance(value, int) or isinstance(value, float):
-            return value
-        if isinstance(value, str):
-            return value
-        if isinstance(value, dt.datetime):
-            return value.isoformat()
-        if isinstance(value, list):
-            return (Pagination._serialize_item(i) for i in value)
-        if isinstance(value, dict):
-            return {Pagination._serialize_item(k): Pagination._serialize_item(v)
-                    for k, v in iteritems(value)}
-        msg = "Unable to serialize {}".format(value)
-        raise TypeError(msg)
+    def _row_proxy_to_dict(row_proxy):
+        """Converts a sqlalchemy row proxy into a vanilla python dictionary
+        were the key is the column name and the value is the value for that
+        record. This method handles the case in which the model defines a
+        different column name than the one defined in the model.
+        """
+        d = {}
+        for column in row_proxy.__table__.columns:
+            class_attr = row_proxy.__mapper__.get_property_by_column(column).key
+            d[column.name] = getattr(row_proxy, class_attr)
+        return d
 
-    @staticmethod
-    def _serialize_row(row_proxy):
-        """Converts each row proxy into a dict and serializes the values"""
-        return {k: Pagination._serialize_item(v)
-                for k, v in iteritems(dict(row_proxy))}
+    def _convert_items_to_dict(self):
+        """Converts each row in the collection of items into a dictionary.
 
-    def serialize_items(self):
-        """JSON serialization of the items"""
-        return (Pagination._serialize_row(row) for row in self.items)
+        A RowProxy object by default is not serializable. This method is used
+        to convert each row to a dictionary of key, value pairs where the key
+        is the column name and the value is the stored value in the database.
+        Note that some datatypes returned in the :class:`Pagination` object
+        are not serializable by default.
+        """
+        return (Pagination._row_proxy_to_dict(row) for row in self.items)
 
     def as_dict(self):
-        """Returns the pagination object as a dict for JSON-like responses"""
+        """Returns the :class: `Pagination` object as a dict for JSON-like responses
+
+        Not all items that are returned from SQLAlchemy's result set is
+        serializable by default although we provide this as a convience method,
+        it is up to the developer to ensure that the correct serialization
+        scheme can be applied to all items that are queried from the database.
+        """
         return {
-            'items': self.serialize_items(),
+            'items': self._convert_items_to_dict(),
             'page': self.page,
             'total': self.total,
             'prev_num': self.prev_num,
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import
 
-import datetime as dt
 import functools
 import os
 import sys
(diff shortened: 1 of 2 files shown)


flask_sqlalchemy/__init__.py
```diff
@@ -372,6 +373,47 @@ def next_num(self):
             return None
         return self.page + 1
 
+    @staticmethod
+    def _serialize_item(value):
+        """Handles specific JSON serialization of different datatypes"""
+        if isinstance(value, bool):
+            if value:
+                return 'true'
```
jab: nit: `return str(value).lower()` handles both true and false at the same time

flask_sqlalchemy/__init__.py
```diff
@@ -372,6 +373,47 @@ def next_num(self):
...
+            return value
+        if isinstance(value, str):
+            return value
+        if isinstance(value, dt.datetime):
+            return value.isoformat()
+        if isinstance(value, list):
+            return [Pagination._serialize_item(i) for i in value]
+        if isinstance(value, dict):
+            return {Pagination._serialize_item(k): Pagination._serialize_item(v)
+                    for k, v in value.items()}
```
jab: nit: use the `iteritems` from _compat.py to avoid wasting resources in Python 2)

flask_sqlalchemy/__init__.py
```diff
@@ -372,6 +373,47 @@ def next_num(self):
...
+        msg = "Unable to serialize {}".format(value)
+        raise NotImplementedError(msg)
+
+    @staticmethod
+    def _serialize_row(row_proxy):
+        """Converts each row proxy into a dict and serializes the values"""
+        return {k: Pagination._serialize_item(v)
+                for k, v in dict(row_proxy).items()}
```
jab: iteritems again

tests/test_pagination.py
```diff
@@ -27,6 +29,43 @@ def test_pagination_pages_when_total_is_none():
...
+    }
+
+
+def test_serialize_row_item_method():
+    test_date = dt.datetime(year=1970, month=1, day=1, hour=12, minute=12, second=12)
+    assert fsa.Pagination._serialize_item(True) == 'true'
+    assert fsa.Pagination._serialize_item(False) == 'false'
+    assert fsa.Pagination._serialize_item(5) == 5
+    assert fsa.Pagination._serialize_item(.1) == .1
+    assert fsa.Pagination._serialize_item([1,'a', .1, test_date]) == [1, 'a', .1, '1970-01-01T12:12:12']
```
jab: nit: space after comma

tests/test_pagination.py
```diff
@@ -27,6 +29,43 @@ def test_pagination_pages_when_total_is_none():
...
+    assert p.as_dict() == {
+        'items': p.items,
+        'page': p.page,
+        'total': p.total,
+        'prev_num': p.prev_num,
+        'next_num': p.next_num
+    }
+
+
+def test_serialize_row_item_method():
```
jab: Worth testing passing in a dict to serialize_item for completeness?

tests/test_pagination.py
```diff
@@ -27,6 +29,43 @@ def test_pagination_pages_when_total_is_none():
...
+    test_date = dt.datetime(year=1970, month=1, day=1, hour=12, minute=12, second=12)
```
jab: Should we add a test for a datetime with a timezone?

flask_sqlalchemy/__init__.py
```diff
@@ -372,6 +373,47 @@ def next_num(self):
...
+
+    def serialize_items(self):
+        """JSON serialization of the items"""
+        return [Pagination._serialize_row(row) for row in self.items]
```
jab: Maybe dumb q but just in case: Should this return a generator comprehension?

flask_sqlalchemy/__init__.py
```diff
@@ -372,6 +373,47 @@ def next_num(self):
...
+        if isinstance(value, dict):
+            return {Pagination._serialize_item(k): Pagination._serialize_item(v)
+                    for k, v in value.items()}
+        msg = "Unable to serialize {}".format(value)
+        raise NotImplementedError(msg)
+
+    @staticmethod
+    def _serialize_row(row_proxy):
+        """Converts each row proxy into a dict and serializes the values"""
+        return {k: Pagination._serialize_item(v)
```
jab: maybe dumb again but: should this yield (k, v) pairs rather than building up the entire dict before returning?

flask_sqlalchemy/__init__.py
```diff
@@ -372,6 +373,47 @@ def next_num(self):
...
+            return value
+        if isinstance(value, dt.datetime):
+            return value.isoformat()
+        if isinstance(value, list):
+            return [Pagination._serialize_item(i) for i in value]
+        if isinstance(value, dict):
+            return {Pagination._serialize_item(k): Pagination._serialize_item(v)
+                    for k, v in value.items()}
+        msg = "Unable to serialize {}".format(value)
+        raise NotImplementedError(msg)
```
jab: `TypeError`?

tests/test_pagination.py
jab: Also, passing in an unhandled type and asserting we get the expected error

rtilk89:
Hi @jab thanks for your feedback! 

I've made the changes you requested except:
* The dictionary generator - do you mean something like:
```python
def _serialize_row(row):
    return dict((k, serialize(v)) for k, v in iteritems(row))
```
* Testing a tz-aware datetime. Gave it a quick shot, but might involve doing some six magic to get it to work since it seemed timezone objects are new in 3.6. If the maintainers think it's important to have in the test suite, I can put a little more effort in

In terms of the project as a whole, I amended the commit and force pushed to only keep one commit in the history. Is it okay to have multiple commits? Or should you rebase to only keep 1 commit?

davidism:
I think my comments in #485 were unclear. I didn't mean we should start serializing rows, I meant we don't want to start doing that. It should be up to the JSON serializer, such as Flask's `jsonify`, to handle whatever's thrown at it. I was worried though that by adding this method we were implying that the data it returned was ready for JSON, which is not true.

I don't want to add row serialization because it's basically a reimplementation of `json.dumps` but not extensible in the same way. The method in this PR converts everything to a string, which will then make everything look like strings when passed through `json.dumps`, rather than the correct JS types. The problem isn't with these data types, which Flask's `jsonify` and `json.dumps` will already handle, it's with queries like `db.session.query(User)`, where the result is a list of `User` objects that aren't serializable.

rtilk89:
@davidism Thanks for your response. The collection of rowProxies (self.items) are not ready to be serialized as is (your comment in #485). 

Should we simply call dict on each item in the collection to turn each row into a dict of <column_name>: <value> pairs that could be fed into json.dumps (or some other encoder)? 

Or your comment that this method may not be useful mean that we should close #482 as it's not worth implementing?

davidism:
Hi, thanks for working on this. I'm going to close it as I no longer think this is a feature we want. I appreciate all the work you did on it, I'm really sorry this is the conclusion. Just having this submitted and seeing the different implementation ideas was the prompting I needed to evaluate and make a decision, so it was helpful!

PR #765, opened 2019-07-23
Add Dyamic Filtering

vgsantoniazzi:
Hi mates.

I've been working with filters for a while and always needed to use an utils to do the job. For example:
```py
@app.route('/places')
def places():
    places = Place.dynamic_filter(filters()).all()
    return render_template('places.html', places=places)

def filters():
    filters = [
        (Place.search_id, request.args.get('search_id_operation') or 'eq', request.args.get('search_id'))
    ]
    return [(key, op, value) for (key, op, value) in filters if value]
```

I was thinking that it will be nice to have in `flask-sqlalchemy` an util like this one.

```py
Todo.query.dynamic_filter([(Todo.title, 'eq', 'Test'), (Todo.text, 'eq', 'test_text')]).all()
```

What do you people think?

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -415,6 +415,37 @@ class BaseQuery(orm.Query):
     This is the default :attr:`~Model.query` object used for models, and exposed as :attr:`~SQLAlchemy.Query`.
     Override the query class for an individual model by subclassing this and setting :attr:`~Model.query_class`.
     """
+    def dynamic_filter(self, filter_conditions, filters = None):
+        '''
+        Return filtered queryset based on condition.
+        :param query: takes query
+        :param filter_condition: Its a list, ie: [(key,operator,value)]
+        operator list:
+            eq for ==
+            lt for <
+            ge for >=
+            in for in_
+            like for like
+            value could be list or a string
+        :return: queryset
+        '''
+        if len(filter_conditions) == 0:
+            return filters
+        else:
+            column, op, value = filter_conditions.pop()
+            try:
+                attr = list(filter(lambda e: hasattr(column, e % op), ['%s', '%s_', '__%s__']))[0] % op
+            except IndexError:
+                raise Exception('Invalid filter operator: %s' % op)
+            if value == 'null':
+                value = None
+            filt = getattr(column, attr)(value)
+            if filters == None:
+                filters = self.filter(filt)
+            else:
+                filters = filters.filter(filt)
+            return self.dynamic_filter(filter_conditions, filters)
+
 
     def get_or_404(self, ident, description=None):
         """Like :meth:`get` but aborts with 404 if not found instead of returning ``None``."""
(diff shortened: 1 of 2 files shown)


ThiefMaster:
I don't see how this needs to be in the core - that could be easily put in your application or a standalone library...

ThiefMaster:
Also, SQLAlchemy already lets you do this - nothing stops you from constructing a filter like `Todo.title == 'test'` from some other data structure

vgsantoniazzi:
@ThiefMaster I was thinking that would be nice, like we have 

```py
    def get_or_404(self, ident, description=None):

    def first_or_404(self, description=None):

    def paginate(self, page=None, per_page=None, error_out=True, max_per_page=None, count=True):
```

Would be nice have dynamic filtering.

Yes, we can do `Test.query.filter(Todo.title == 'test')` but it does not work very well if we're building an application with those filters optional, and we will need to have an structure like above to work:

```
filters = Todo.query.filter()

if request.args.get('search_id'):
    filters = filters.filter(Todo.search_id == request.args.get('search_id'))
```

Something like

```
Todo.query.filter(Todo.search_id == request.args.get('search_id'))
```

Will not work with optional filtering

What do you think?

ThiefMaster:
Still: You can implement the logic in your application :)

vgsantoniazzi:
@ThiefMaster yes, Like the pagination. I have a lot of projects that I need the same source code for filtering.

If do you think it's a good idea, let me know.

Best

vgsantoniazzi:
If someone whats to implement dynamic filtering, you can use this example here:

```py
    @classmethod
    def dynamic_filter(model_class, filter_condition):
        '''
        Return filtered queryset based on condition.
        :param query: takes query
        :param filter_condition: Its a list, ie: [(key,operator,value)]
        operator list:
            eq for ==
            lt for <
            ge for >=
            in for in_
            like for like
            value could be list or a string
        :return: queryset
        '''
        __query = db.session.query(model_class)
        for raw in filter_condition:
            try:
                key, op, value = raw
            except ValueError:
                raise Exception('Invalid filter: %s' % raw)
            column = getattr(model_class, key, None)
            if not column:
                raise Exception('Invalid filter column: %s' % key)
            if op == 'in':
                if isinstance(value, list):
                    filt = column.in_(value)
                else:
                    filt = column.in_(value.split(','))
            else:
                try:
                    attr = list(filter(lambda e: hasattr(column, e % op), ['%s', '%s_', '__%s__']))[0] % op
                except IndexError:
                    raise Exception('Invalid filter operator: %s' % op)
                if value == 'null':
                    value = None
                filt = getattr(column, attr)(value)
            __query = __query.filter(filt)
        return __query
```

and in controller

```py
@app.route('/places')
def places():
    places = Place.dinamic_filter(filters()).all()
    return render_template('places.html', places=places)


def filters():
    filters = [
        ('search_id', request.args.get('search_id_operation') or 'eq', request.args.get('search_id'))
    ]
    return [(key, op, value) for (key, op, value) in filters if value]
```
```

PR #771, opened 2019-08-20
Use updated interface for SQLAlchemy 1.3

mpessas:
See https://github.com/sqlalchemy/sqlalchemy/issues/4586 and https://docs.sqlalchemy.org/en/13/orm/mapping_api.html#sqlalchemy.orm.mapper.Mapper.mapped_table.


mpessas:
Sorry, wrong repo.

PR #780, opened 2019-09-02
Fix TimeoutError when QueuePool limit is reached

welisonmenezes:
I got the Error: 'sqlalchemy.exc.TimeoutError: QueuePool limit of size 10 overflow 10 reached, connection timed out, timeout 30'.

To solve this problem I changed line 852: self.session.remove() by
db.session.close_all().


davidism:
This would close other sessions for active request in the same process.

PR #791, opened 2019-10-24
Add docs and example of using plain SQLAlchemy models with Flask-SQLAlchemy

e271828-:
There are many examples of people struggling to work around this limitation in flask-sqlalchemy, including various issues and PRs like:

https://github.com/pallets/flask-sqlalchemy/issues/269 
https://github.com/pallets/flask-sqlalchemy/pull/250
https://github.com/pallets/flask-sqlalchemy/pull/629

This PR documents a simple solution to bi-directional compatibility.

### DIFF
--- a/docs/models.rst
+++ b/docs/models.rst
@@ -167,3 +167,82 @@ a query object using
 :meth:`Page.query.with_parent(some_tag) <sqlalchemy.orm.query.Query.with_parent>`
 and then use it exactly as you would with the query object from a dynamic
 relationship.
+
+Using Plain SQLAlchemy Models
+-----------------------------
+
+You may have pre-existing SQLAlchemy models that you would like to use
+with Flask-SQLAlchemy, or conversely you may find yourself needing to access
+the Flask-SQLAlchemy models you have already defined outside of a Flask context.
+
+The simplest method is to simply switch the base used for your models at run-time,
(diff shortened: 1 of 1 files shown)


docs/models.rst
```diff
@@ -167,3 +167,81 @@ a query object using
...
+A complete worked example looks something like this::
+
+	import os
+	from sqlalchemy import create_engine, Column, String, Integer
+	from sqlalchemy.ext.declarative import declarative_base
+	from sqlalchemy.orm import scoped_session, sessionmaker
+	from flask_sqlalchemy import SQLAlchemy
+
+	# We can control this via `PLAIN_SQLALCHEMY_BASE=true python3 external_use.py`
+	PLAIN_SQLALCHEMY_BASE = "true" in os.getenv("PLAIN_SQLALCHEMY_BASE", "false").lower()
```
ThiefMaster: I don't think runtime switching makes much sense for something like this...

e271828-: > I don't think runtime switching makes much sense for something like this...

The obvious alternative is refactoring initialization to make `model_class` apply to generic models instead of only working properly (e.g. attaching query class) when flask_sqlalchemy.Model is subclassed.

Seems like people have been proposing patches to do that for several years, none of which have landed. 

What would you suggest?

davidism:
Closing for similar reasons to #629.

PR #804, opened 2020-02-18
Removed conditional count from pagination

CoburnJoe:
If you have duplicate results (which get deduped), `len(items)` always returns a value less than the per_page limit. This causes a bug whereby you have several pages of results, but the pagination metadata does not reveal this, and implies there is only one page of results.

The overhead from counting all results now, regardless of whether they fit onto a single page is worth considering, but using the `count ` flag can bypass this as before.

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -499,13 +499,8 @@ class BaseQuery(orm.Query):
         if not items and page != 1 and error_out:
             abort(404)
 
-        # No need to count if we're on the first page and there are fewer
-        # items than we expected or if count is disabled.
-
         if not count:
             total = None
-        elif page == 1 and len(items) < per_page:
-            total = len(items)
         else:
             total = self.order_by(None).count()
 


nomad-software:
@davidism Does this look good to you?

davidism:
Now that you bring this up, does `query.count()` dedupe items, or is that a behavior later within SQLAlchemy?

CoburnJoe:
No, it's getting de-duped inside ''.all()'', where ''limit'' and ''offset'' get applied to the query just before counting the results.

davidism:
As I understand it, `count` is *instead of* `all`, and happens on the SQL side, where SQLAlchemy doesn't do any deduping.

CoburnJoe:
This PR addresses a bug whereby results on the _first_ page get counted in a totally different way to any other pages of results (which breaks the paging logic if there are fewer results than expected, e.g. because of duplicates). 

I don't see why SQL itself would dedupe the results (although it is possible, as I could not replicate this issue with SQLite), but I think that's a different conversation around if de-duping should happen at all (which probably isn't the fault of flask-sqlalchemy).

davidism:
My point was that `len(items)` is more accurate than `query.count()` because it takes into account what will actually be returned. Pagination and counting in general need a lot of attention though, I think this is fine for now just to make it consistent at least.

CoburnJoe:
Ahh, gotcha - I've seen some of the confusion and discussion around paging in general - I've got some ideas but I'll not send this off-topic :)

nomad-software:
Merge this mutha! :+1:

RazerM:
@CoburnJoe can you provide an example where duplicates would affect the count?

CoburnJoe:
> @CoburnJoe can you provide an example where duplicates would affect the count?

That's a bit hard to do - and this is now closed and released in 2.4.2.

Anyway, (previously), by counting ```len(items)``` and ```self.order_by(None).count()```, you get two different totals depending on the results returned, which is inconsistent, and a big problem when items gets de-duped, and returns less results than the ```per_page``` threshold.

RazerM:
> That's a bit hard to do - and this is now closed and released in 2.4.2.

Yes, sorry to ask after it's merged but I noticed the additional queries in our snapshot tests after upgrading.

> Anyway, (previously), by counting `len(items)` and `self.order_by(None).count()`, you get two different totals depending on the results returned, which is inconsistent, and a big problem when items gets de-duped, and returns less results than the `per_page` threshold.

I don't believe this to be the case, but maybe I'm missing something. When do items get de-duped?

CoburnJoe:
@RazerM perhaps this would have been cleaner to switch everything over to ```len()``` instead of ```count()``` (if possible to count the correct object) - this bug caused a massive issue for one of our APIs, and I would think issues for many other users, even if they didn't realise.

> I don't believe this to be the case, but maybe I'm missing something. When do items get de-duped?

_Somewhere_ inside SQLAlchemy itself, duplicate results get de-duped. This could be a bad query, or perhaps a bad database design and dodgy data (e.g. a left join on bad data would duplicate records).

Here's an example. Suppose your query returns 250 records. You have ```per_page``` set to 50, so that's five pages of 50 results each.

```count()``` correctly returns 250 - the total number of records. ```len()``` (in _this_ case) incorrectly returns 50 - the number of items on this page. Supposing you have a duplicate deduped. Now you have 49 results - which is under the ```per_page``` threshold. Now, pagination stops as it thinks there are only 49 results. You have no idea other pages exist, even though there are 5 pages of result, you only think there is one.

**Comment by RazerM** [comment]
I did a bit of digging and found this issue: https://github.com/sqlalchemy/sqlalchemy/issues/4395

In short, SQLAlchemy deduplicates ORM queries that return objects, which when combined with a join that duplicates rows causes the problem. It looks like this will be fixed in SQLAlchemy 2. I will submit a PR with a regression test because it should nicely demonstrate why the old pagination behaviour was wrong.

I'm sorry if I wasted your time asking about this, the SQLAlchemy behaviour is surprising here even if it only applies to 'bad' queries.

**Comment by CoburnJoe** [comment]
> I did a bit of digging and found this issue: [sqlalchemy/sqlalchemy#4395](https://github.com/sqlalchemy/sqlalchemy/issues/4395)
> 
> In short, SQLAlchemy deduplicates ORM queries that return objects, which when combined with a join that duplicates rows causes the problem. It looks like this will be fixed in SQLAlchemy 2. I will submit a PR with a regression test because it should nicely demonstrate why the old pagination behaviour was wrong.
> 
> I'm sorry if I wasted your time asking about this, the SQLAlchemy behaviour is surprising here even if it only applies to 'bad' queries.

You didn't waste my time! It's a subtle issue, but one that I spent several days tracking down in my projects.

**Comment by ameyers123** [comment]
Hey all! I've just implemented pagination via flask-sqlalchemy, and i'm running into this error. 

It seems pagination sometimes provides extra entries and does not follow the "items_per_page_limit".

```
all_trips = Trips.query.filter_by(is_listed=True) \
        .filter(and_((Trips.start_date <= end_date) & (Trips.end_date >= start_date))) \
        .filter(not_((Trips.start_date < start_date) & (Trips.end_date > end_date))) \
        .join(Leg, Leg.trip_uuid == Trips.uuid) \
        .join(User, User.uuid == Trips.user_uuid) \
        .filter(*filters) \
        .order_by(*sort_filters)
        
    trips = all_trips.paginate(page_number, items_per_page_limit, page_flag)

    results_total = len(all_trips.all())
    result_count_for_page = len(trips.items)
    pages_total = trips.pages
```

davidism:
Please use Stack Overflow for questions about your own code. This tracker is for issues related to the project itself. Be sure to include a [minimal, complete, and verifiable example](https://stackoverflow.com/help/mcve).

PR #805, opened 2020-02-23
Remove default pool options for MySQL NullPools

### LINKED ISSUE #803: Issue using NullPool poolclass in MySQL
I am using version 2.4.1. I am trying to set my poolclass to NullPool by passing that option into my engine options like so:
```
from sqlalchemy.pool import NullPool
SQLALCHEMY_ENGINE_OPTIONS = {"poolclass":NullPool}
```
That is working as expected and the poolclass is being set. However I am getting the following error when running:
```
TypeError: Invalid argument(s) 'pool_size' sent to create_engine(), using configuration AuroraMySQLDataAPIDialect/NullPool/Engine.  Please check that the keyword arguments are appropriate for this combination of components.
```

The pool_size needs to NOT be set in order to use NullPool. This seems to be happening because the pool_size is being set to 10. I did some digging and found it was being set in apply_driver_hacks:
```
        if sa_url.drivername.startswith('mysql'):
            sa_url.query.setdefault('charset', 'utf8')
            if sa_url.drivername != 'mysql+gaerdbms':
                options.setdefault('pool_size', 10)
                options.setdefault('pool_recycle', 7200)
```
If I comment out those last 3 lines and run my code, the pool size doesn't get set and everything works as expected.

But I would like to figure out a way to prevent the pool size from being set to 10 without having to fork the repo and make this change to do so. Is there any way to do this? I have tried passing in pool_size:0 and pool_size:None into my engine options but it is still getting set to 10 regardless.

FWIW I am using a custom dialect (https://github.com/chanzuckerberg/sqlalchemy-aurora-data-api), but the driver name starts with "mysql" so it's being caught by this logic regardless.

CoburnJoe:
This proposed change fixes #803 

As mentioned in the issue, if you use MySQL dialects (or custom dialects starting with "MySQL") with a NullPool, the old logic triggers and attempts to apply pool defaults to a NullPool, which subsequently fails.

By checking ``options['pool_class']`` first, skip the logic for ``NullPool`` classes.

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -23,7 +23,6 @@ iniconfig==1.0.0          # via pytest
 jinja2==2.11.2            # via sphinx
 markupsafe==1.1.1         # via jinja2
 mock==4.0.2               # via -r requirements/tests.in
-more-itertools==8.3.0     # via pytest
 nodeenv==1.3.5            # via pre-commit
 packaging==20.4           # via pallets-sphinx-themes, pytest, sphinx, tox
 pallets-sphinx-themes==1.2.3  # via -r requirements/docs.in
@@ -33,7 +32,7 @@ pre-commit==2.7.1         # via -r requirements/dev.in
 py==1.9.0                 # via pytest, tox
 pygments==2.6.1           # via sphinx
 pyparsing==2.4.7          # via packaging
-pytest==6.0.2             # via -r requirements/tests.in
+pytest==6.1.0             # via -r requirements/tests.in
 pytz==2020.1              # via babel
 pyyaml==5.3.1             # via pre-commit
 requests==2.23.0          # via sphinx
--- a/requirements/tests.txt
+++ b/requirements/tests.txt
@@ -8,11 +8,10 @@ attrs==19.3.0             # via pytest
 blinker==1.4              # via -r requirements/tests.in
 iniconfig==1.0.0          # via pytest
 mock==4.0.2               # via -r requirements/tests.in
-more-itertools==8.3.0     # via pytest
 packaging==20.4           # via pytest
 pluggy==0.13.1            # via pytest
 py==1.9.0                 # via pytest
 pyparsing==2.4.7          # via packaging
-pytest==6.0.2             # via -r requirements/tests.in
+pytest==6.1.0             # via -r requirements/tests.in
 six==1.15.0               # via packaging
 toml==0.10.1              # via pytest


flask_sqlalchemy/__init__.py
```diff
@@ -874,8 +874,14 @@ def apply_driver_hacks(self, app, sa_url, options):
         if sa_url.drivername.startswith('mysql'):
             sa_url.query.setdefault('charset', 'utf8')
             if sa_url.drivername != 'mysql+gaerdbms':
-                options.setdefault('pool_size', 10)
-                options.setdefault('pool_recycle', 7200)
+                try:
+                    pool_class = options['poolclass']
+                except KeyError:
+                    pool_class = None
```
thestub: Would it be better to do 

`if 'poolclass' in options:`

rather than having a side effect behaviour?

CoburnJoe: I'd define this as defensive programming rather than a side effect, but yes, your suggestion is far better :)

CoburnJoe: Updated

thestub: I wouldn't go with far better :flushed: 😳

ThiefMaster: What about `pool_class = options.get('poolclass')`?

ThiefMaster: ah nvm just realized the code the comments were on doesn't even exist anymore

thestub: > What about `pool_class = options.get('poolclass')`?

I like that. A much less expensive operation.
Would it be good to explicitly add a default to make obvious that we selected a `get` to provide a `None`? Although we test for that a couple of lines later - it should be obvious.

ThiefMaster: No, returning `None` is literally the default of `dict.get()` in the stdlib.

CoburnJoe: Same efficiency, but could still reduce the line count from 3 -> 1

thestub: @ThiefMaster You can tell I have been bitten by idiots in the past 😂

ThiefMaster: I usually expect people to know the basics of the language they are using ;)

thestub: Too often that's proven to not be the case 😔

CoburnJoe: `options.get` is a nice choice, but you still need to handle None types to retrieve `__name__`, so I don't think there would be any significant benefits over the (now revised) logic.

thestub: Shouldn't we then be setting the name to `NullType`

thestub: Ignore that. Being simple.

flask_sqlalchemy/__init__.py
```diff
@@ -874,8 +874,13 @@ def apply_driver_hacks(self, app, sa_url, options):
         if sa_url.drivername.startswith('mysql'):
             sa_url.query.setdefault('charset', 'utf8')
             if sa_url.drivername != 'mysql+gaerdbms':
-                options.setdefault('pool_size', 10)
-                options.setdefault('pool_recycle', 7200)
+                pool_class_name = None
+                if 'poolclass' in options:
+                    pool_class_name = pool_class.__name__
```
thestub: Where is `pool_class` coming from? It was being loaded from the dict.

flask_sqlalchemy/__init__.py
```diff
@@ -874,8 +874,13 @@ def apply_driver_hacks(self, app, sa_url, options):
...
+                if pool_class_name != 'NullPool':
```
thestub: I think this still needs to check that the key exists, as per your original commit.

thestub:
Should the change actually be

```
pool_class = options.get('poolclass') 
if pool_class and pool_class.__name__ != 'NullPool':
    # NullPool doesn't support pooling options...
    options.setdefault('pool_size', 10)
    options.setdefault('pool_recycle', 7200)`
```

CoburnJoe:
> Should the change actually be
> 
> ```
> pool_class = options.get('poolclass') 
> if pool_class and pool_class.__name__ != 'NullPool':
>     # NullPool doesn't support pooling options...
>     options.setdefault('pool_size', 10)
>     options.setdefault('pool_recycle', 7200)`
> ```

No - that would give you an `AttributeError` when there is no poolclass.

flask_sqlalchemy/__init__.py
CoburnJoe: Yes good spot - fixed

flask_sqlalchemy/__init__.py
```diff
@@ -874,8 +874,13 @@ def apply_driver_hacks(self, app, sa_url, options):
         if sa_url.drivername.startswith('mysql'):
             sa_url.query.setdefault('charset', 'utf8')
             if sa_url.drivername != 'mysql+gaerdbms':
-                options.setdefault('pool_size', 10)
-                options.setdefault('pool_recycle', 7200)
+                pool_class_name = None
+                if 'poolclass' in options:
```
davidism: `pool_class` needs an underscore here.

CoburnJoe: If we scroll down a few lines there are examples of no underscore: `options['poolclass'] = StaticPool`.

I appreciate this PR has mixed both underscores and no underscores, so my question is, which should it be, and does the rest of the code need fixing?

davidism: Never mind, it's the name of the argument in SQLAlchemy: https://docs.sqlalchemy.org/en/13/core/engines.html#sqlalchemy.create_engine.params.poolclass.

Which means the next line needs to be `options['poolclass']`.

thestub:
Just to double check, we only want to not set the pool defaults if the `pool_class` is NullPool? Currently it will attempt to set the defaults if there is no `pool_class` key in the dict.

flask_sqlalchemy/__init__.py
CoburnJoe: Done :)

olinger:
Any update on getting this merged? I am stuck using a fork of flask-sqlalchemy in my project to get around this.

bfaivre:
Checking in on this PR status. My project could really benefit from using the NullPool class w/MySQL. Thanks in advance.

CoburnJoe:
@bfaivre there's still some queries around this change, and it now has conflicts - give me a few days to rework it and we'll take it from there.

bfaivre:
@CoburnJoe do you know of a current work around? My application is only using the NullPool class so I've just commend out the following in __init__.py:

```
# if sa_url.drivername != 'mysql+gaerdbms':
#     options.setdefault('pool_size', 10)
#     options.setdefault('pool_recycle', 7200)

```

However if there is a more appropriate work around I'd rather use that for the time being vs. modifying the code. Thanks again.

davidism:
It's using setdefault, so no need to patch out library code, just pass those keys with None for the value.

bfaivre:
> It's using setdefault, so no need to patch out library code, just pass those keys with None for the value.

Thanks for the quick response @davidism. Just to confirm with you, this is still an issue even w/passing None using setdefault and presenting the following error:

`TypeError: Invalid argument(s) 'pool_size' sent to create_engine(), using configuration MySQLDialect_mysqldb/NullPool/Engine.`

Just wanting to make sure I'm not missing anything obvious here. Thanks again for all your great work.

Cheers,

Brian

davidism:
looks like something went wrong here, there's no changes listed

CoburnJoe:
> looks like something went wrong here, there's no changes listed

Yep - I closed off as it's a bit...broken!
I've discussed in #803 - I'm going to try revisit this in the next week

Issue #567, opened 2017-11-17
Add "first" and "last" to Pagination class?


jessesheidlower:
I'm new to Flask, and was wondering about a feature that I think would be useful to add to the Pagination class. Unless I'm missing it, there doesn't seem to be a built-in way to get the numbers of the items you're viewing on the page itself. That is, if you want to display "387 records found; displaying 26–50", you'd have to do calculations to get the "26" and "50".

I think something along these lines would work:

```
@property
    def first(self):
        """The number of the first item on the current page"""
        if self.total == 0:
            first = 0
        else:
            first = ((self.page - 1) * self.per_page) + 1
        return first

  @property
    def last(self):
        """The number of the last item on the current page"""
        if self.page == self.pages:
            last = self.total
        else:
            last = self.page * self.per_page
        return last
```
Then you could do something like (in jinja2):

   ```{{ pagination.total }} records found; displaying {{ pagination.first }} – {{pagination.last }}```

I'm new to Python programming, and I don't know enough about testing, running changed code, etc., to feel comfortable submitting this as a pull request, so I hope it's OK that I just posted this as an Issue.

Thanks.

rsyring:
The PR for this creates a problem in our test suite.  See the latest comment there if interested.

PR #586, opened 2018-01-20
Implemented first and last property for pagination

asoderman:
Added first and last property to the pagination class and relevant tests.

Relevant issue: #567 

Travis failed because of python 2.6 no longer being supported by SQLAlchemy see #587.

PR #810, opened 2020-03-26
Added first & last property and also items_length attribute to the pagination

mwxgaf:
#567 
first and last property added because of the mentioned issue.(you can find more info in the issue)

Added items_length attribute which shows the number of items for current page that is useful for showing the number of items in a page specially the last one if not total % per_page == 0.


davidism:
Duplicate of #586

PR #811, opened 2020-03-30
missing dependency when deploying to a clean environment.

aaronwolmutt:


ThiefMaster:
While postgres is great, you can use flask-sqlalchemy/sqlalchemy with many different databases and database client libraries. So it's completely up to you which one to install.

PR #866, opened 2020-08-05
add more detailed description for create_all

zachvalenta:
just added line from SQLAlchemy docs ([link](https://docs.sqlalchemy.org/en/13/core/metadata.html#sqlalchemy.schema.MetaData.create_all)) that seems germane i.e. "create_all(), idempotent, yay or nay?"

### DIFF
--- a/src/flask_sqlalchemy/__init__.py
+++ b/src/flask_sqlalchemy/__init__.py
@@ -1041,26 +1041,40 @@ class SQLAlchemy:
             op(bind=self.get_engine(app, bind), **extra)
 
     def create_all(self, bind="__all__", app=None):
-        """Creates all tables.
+        """Create all tables that do not already exist in the database.
+        This does not update existing tables, use a migration library
+        for that.
+
+        :param bind: A bind key or list of keys to create the tables
+            for. Defaults to all binds.
+        :param app: Use this app instead of requiring an app context.
 
         .. versionchanged:: 0.12
-           Parameters were added
+            Added the ``bind`` and ``app`` parameters.
         """
         self._execute_for_all_tables(app, bind, "create_all")
 
     def drop_all(self, bind="__all__", app=None):
-        """Drops all tables.
+        """Drop all tables.
+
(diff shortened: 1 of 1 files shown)


CoburnJoe:
Thanks for your contribution! Unfortunately, this change is failing the style guide section of the build: https://github.com/pallets/flask-sqlalchemy/runs/951073509

The [Contribution guide](https://github.com/pallets/flask-sqlalchemy/blob/master/CONTRIBUTING.rst) covers the pre-commit hooks and other tools and requirements to get this to pass.

zachvalenta:
adjusted line length, seems like it's passing CI now?

src/flask_sqlalchemy/__init__.py
```diff
@@ -999,7 +999,11 @@ def _execute_for_all_tables(self, app, bind, operation, skip_tables=False):
             op(bind=self.get_engine(app, bind), **extra)
 
     def create_all(self, bind="__all__", app=None):
-        """Creates all tables.
+        """
+            Creates all tables.
```
davidism, line 1003: First line should be on the same line as the `"""`. Additional lines should not be indented.

zachvalenta: my bad, got that format from [here](https://github.com/pallets/flask-sqlalchemy/blob/master/src/flask_sqlalchemy/__init__.py#L934): 

```python
# __init.py__
    def create_engine(self, sa_url, engine_opts):
        """
            Override this method to have final say over how the SQLAlchemy engine
            is created.
            In most cases, you will want to use ``'SQLALCHEMY_ENGINE_OPTIONS'``
            config variable or set ``engine_options`` for :func:`SQLAlchemy`.
        """
        return sqlalchemy.create_engine(sa_url, **engine_opts)
```

figured it might be a way to handle more verbose docstrings that had more than one line.

zachvalenta: just updated and re-pushed

PR #880, opened 2020-10-24
init_app now sets  the app attribute

italomaia:
Now calling init_app after constructing the flask extension behaves the same
as providing app to the constructor.
Also, setting the app attribute on the constructor only creates problems for the **empty** package and flask-empty boilerplate. If not setting app on **init_app** was by design and has a technical justification, please, let me know.

### DIFF
--- a/src/flask_sqlalchemy/__init__.py
+++ b/src/flask_sqlalchemy/__init__.py
@@ -727,7 +727,6 @@ class SQLAlchemy:
         self.session = self.create_scoped_session(session_options)
         self.Model = self.make_declarative_base(model_class, metadata)
         self._engine_lock = Lock()
-        self.app = app
         self._engine_options = engine_options or {}
         _include_sqlalchemy(self, query_class)
 
@@ -814,6 +813,8 @@ class SQLAlchemy:
         of an application not initialized that way or connections will
         leak.
         """
+
+        self.app = app
 
         # We intentionally don't set self.app = app, to support multiple
         # applications. If the app is passed in the constructor,
--- a/tests/test_extension_class.py
+++ b/tests/test_extension_class.py
@@ -1,0 +1,14 @@
+from flask_sqlalchemy import SQLAlchemy
+
+
+def test_constructor_sets_app(app):
+    db = SQLAlchemy(app)
+
+    assert db.app is not None
+
+
+def test_init_app_sets_app(app):
+    db = SQLAlchemy()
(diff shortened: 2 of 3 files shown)


davidism:
This is incorrect. The current behavior is correct. See lots of previous discussions on this. Push an app context if you want to use the extension outside a view/cli command.

italomaia:
I will look into previous discussions but there are other issues with this behavior; imagine I create my extension instance in a module and use it from from everywhere else. I cannot properly initialize it with **init_app** because init_app creates a new flask extension. So, you use the constructor to explicitly create a flask extension instance then you initialize it, which creates another instance implicitly, and makes your instance completely useless, forcing me to use the new instance in app.extensions, making the whole app design harder.

ThiefMaster:
I think you misunderstood something.

You put `db = SQLAlchemy()` somewhere (let's say `yourapp/db.py`) and then you do `from yourapp.db import db` whenever you need it, e.g. to create a new model.

In you app factory you then import it as well and call `db.init_app(app`). And since a Flask app should never "do stuff" (e.g. accessing the DB) outside an app context, `db.session.query` etc. will work just fine whenever you *are* inside an app context.

italomaia:
@davidism also the snippet in SQLAlchemy init_app doesn't work. 

```
db = SQLAlchemy()
def create_app():
    app = Flask(__name__)
    db.init_app(app)
    return app
```

Because init_app will create a new instance of SQLAlchemy, so db is actually not initialized.
See line https://github.com/pallets/flask-sqlalchemy/blob/master/src/flask_sqlalchemy/__init__.py#L631

ThiefMaster:
> Because init_app will create a new instance of SQLAlchemy, so db is actually not initialized.

This is wrong. are you mixing up `SQLAlchemy` and `_SQLAlchemyState`?

italomaia:
@ThiefMaster you're correct, I mixed SQLAlchemy with _SQLAlchemyState. Let me check the discord channel.

italomaia:
I tried to cook up an example using the app_context to actually try to understand how useful using app_context here would be. Ended up with this: https://pastebin.com/2kjrRyZz 
That can be accomplish using the old style as well, in my opinion, in a more readable way. That said, is there an actual example where requiring app_context to use db methods helps? I am at lost here. I understand this was a design decision and it is not my goal to challenge that. I just want to understand how this choice helps.

PR #886, opened 2020-11-15
Update config.rst

joshua-stauffer:
The current suggested Metadata naming_convention dictionary runs into issues on migrating Boolean columns in SQLite databases as described here (https://github.com/sqlalchemy/sqlalchemy/issues/3345) and here (https://github.com/sqlalchemy/sqlalchemy/issues/4784). If for some reason the original ck value is necessary, it might be useful to have an additional note indicating this as an alternative. Thanks for considering!

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -171,7 +171,7 @@ example, as suggested by the SQLAlchemy docs::
     convention = {
         "ix": 'ix_%(column_0_label)s',
         "uq": "uq_%(table_name)s_%(column_0_name)s",
-        "ck": "ck_%(table_name)s_%(constraint_name)s",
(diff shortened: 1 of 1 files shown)


davidism:
In SQLAlchemy 1.4, non-native Boolean and Enum no longer generate constraints by default since it was causing more issues than it was solving. Additionally, both types already took an optional name to be used for the constraint if necessary.

* https://docs.sqlalchemy.org/en/14/changelog/migration_14.html#enum-and-boolean-datatypes-no-longer-default-to-create-constraint
* https://docs.sqlalchemy.org/en/14/core/type_basics.html#sqlalchemy.types.Boolean

Based on the discussions you linked, it seems more correct to leave this as-is. Projects should use the features mentioned above.

PR #919, opened 2021-02-27
Initial Support for SQLAlchemy 1.4

aadel:
This is an initial PR to support SQLAlchemy 1.4. Since SQLAlchemy 1.4 is in _pre-release_ state at this time, the PR is initial until 1.4 is formally released. In this PR, there are no interface changes and support for SQLAlchemy 1.2+ is preserved. The PR addresses the underlaying SQLAlchemy changes which impacts Flask-SQLAlchemy.

All tests passed against SQLAlchemy `master` branch. Two tests don't pass against 1.4.0b3 though due to an issue with relationship dynamic loading that was fixed on `master` but not released yet:

* tests/test_query_class.py::test_default_query_class
* tests/test_query_class.py::test_custom_query_class

See https://github.com/sqlalchemy/sqlalchemy/issues/5981 for more details.

### Relevant issue

- #910: Update for new versions of SQLAlchemy

### Checklist:

- [X] Add tests that demonstrate the correct behavior of the change. Tests should fail without the change.
- [X] Add or update relevant docs, in the docs folder and in code.
- [X] Add an entry in `CHANGES.rst` summarizing the change and linking to the issue.
- [X] Add `.. versionchanged::` entries in any relevant code docs.
- [X] Run `pre-commit` hooks and fix any issues.
- [X] Run `pytest` and `tox`, no tests failed.

PR #926, opened 2021-03-16
Fix: Limit sqlalchemy dependency to <1.4

### LINKED ISSUE #885: URL is immutable from SQLAlchemy 1.4
### Expected Behavior

Where the SQLAlchemy object is created using the 'second possibility' documented in the SQLAlchemy code, and an app context exists, calling the create_all method in the SQLAlchemy class should be successful.

### Actual Behavior

Because URL objects are [immutable from SQLAlchemy 1.4](https://docs.sqlalchemy.org/en/14/changelog/migration_14.html#the-url-object-is-now-immutable) calling the get_engine method in the _EngineConnector class will cause an AttributeError exception to be thrown due to an attempt to manually modify a previously created URL in the apply_driver_hacks method in the SQLAlchemy class, i.e.

```python
if sa_url.drivername.startswith("mysql"):
            sa_url.query.setdefault("charset", "utf8")
```

I haven't had the chance to test this, but it seems you could replace

```python
  sa_url.query.setdefault("charset", "utf8")
```

with

```python
  sa_url.update_query_dict({"charset" : "utf8"})
```

### Environment

* Python version: 3.8
* Flask-SQLAlchemy version:  2.4.4
* SQLAlchemy version: 1.4.0b1

slicen:
This pull request addresses #885 until #919 is completed.

It ensures that sqlalchemy 1.4 is not pulled in as a dependency prior to adding support for it.

At present, the dependency on sqlalchemy >= 1.2 results in the latest release, 1.4, being added.
The module does not yet support sqlalchemy 1.4, resulting in new installations failing when using
local databases (#885).

- fixes #885

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,7 @@ Unreleased
 
 -   Drop support for Python 2, 3.4, and 3.5.
 -   Bump minimum version of Flask to 1.0.4.
--   Bump minimum version of SQLAlchemy to 1.2.
+-   Bump minimum version of SQLAlchemy to 1.2 and restrict to < 1.4 `885`.
 -   Remove previously deprecated code.
 -   Set ``SQLALCHEMY_TRACK_MODIFICATIONS`` to ``False`` by default.
     :pr:`727`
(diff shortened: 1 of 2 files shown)


davidism:
We don't pin max dependencies in libraries. Projects should use a tool like pip-complie or poetry to control when they upgrade.

PR #931, opened 2021-03-17
Updated CamelCase to snake_case converter to cover more cases

### LINKED ISSUE #406: Acronyms in tablename aren't converted to snake case
Acronyms aren't being recognized as separate words when converting to snake case.  Only happens with plain acronyms or acronyms with a prefix but not a suffix.

``` python
from flask import Flask

app = Flask(__name__)
db = SQLAlchemy(app)

class ABC(db.Model):

class PreABC(db.Model):

class ABCPost(db.Model):

class PreABCPost(db.Model):

db.engine.echo = True
db.create_all()
```

```
CREATE TABLE "ABC" (
    id INTEGER NOT NULL, 
    PRIMARY KEY (id)
)

CREATE TABLE "preABC" (
    id INTEGER NOT NULL, 
    PRIMARY KEY (id)
)

CREATE TABLE abc_post (
    id INTEGER NOT NULL, 
    PRIMARY KEY (id)
)

CREATE TABLE pre_abc_post (
    id INTEGER NOT NULL, 
    PRIMARY KEY (id)
)
```

Saif807380:
I have updated the regex pattern for converting CamelCase table names to snake_case. I've also compared this with #617. Detailed results of the comparison can be found [here](http://bit.ly/test-case-analysis). 

`tests/test_basic_app.py::test_sqlite_relative_path` fails for `SQLAlchemy == 1.4.0` giving #885. All tests pass with older `SQLAlchemy` version.

- fixes #406 

Checklist:

### DIFF
--- a/tests/test_table_name.py
+++ b/tests/test_table_name.py
@@ -4,6 +4,44 @@ import pytest
 from sqlalchemy.exc import ArgumentError
 from sqlalchemy.ext.declarative import declared_attr
 
+from flask_sqlalchemy.model import camel_to_snake_case
+
+
+@pytest.mark.parametrize(
+    ("name", "expect"),
+    [
+        ("CamelCase", "camel_case"),
+        ("Snake_case", "snake_case"),
+        ("HTMLLayout", "html_layout"),
+        ("LayoutHTML", "layout_html"),
+        ("HTTP2Request", "http2_request"),
+        ("ShoppingCartSession", "shopping_cart_session"),
+        ("ABC", "abc"),
+        ("PreABC", "pre_abc"),
+        ("ABCPost", "abc_post"),
+        ("PreABCPost", "pre_abc_post"),
+        ("HTTP2RequestSession", "http2_request_session"),
+        ("UserST4", "user_st4"),
+        (
+            "HTTP2ClientType3EncoderParametersSSE",
+            "http2_client_type3_encoder_parameters_sse",
+        ),
+        (
+            "LONGName4TestingCamelCase2snake_caseXYZ",
+            "long_name4_testing_camel_case2snake_case_xyz",
+        ),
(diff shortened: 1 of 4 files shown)


davidism:
Found a few more test cases I had written out a long time ago. There's probably some overlap with what you tested, but you can add others.

```python
@pytest.mark.parametrize(
    ("name", "expect"),
    [
        ("A", "a"),
        ("A1", "a1"),
        ("1A", "1a"),
        ("CamelCase", "camel_case"),
        ("Snake_Case", "snake_case"),
        ("ABC", "abc"),
        ("HTMLLayout", "html_layout"),
        ("LayoutHTML", "layout_html"),
        ("HTTP2Request", "http2_request"),
        ("4XTest", "4x_test"),
        ("ShoppingCartSession", "shopping_cart_session"),
    ],
)
```

davidism:
Also, use `@pytest.mark.parametrize()` as shown above. This allows each test case to pass/fail individually.

tests/test_table_name.py
```diff
@@ -4,6 +4,54 @@
...
+        "vp9_codec",
+        "http2_request_session",
+        "user_st4",
+        "http2_client_type3_encoder_parameters_sse",
+        "long_name4_testing_camel_case2snake_case_kek",
+        "foo_bar_sse2",
+        "alarm_message_ss2_signal_transformer",
+        "ast_v2_node",
+        "http_response_code_xyz",
+        "get_htt_presponse_code",
```
ThiefMaster, line 45: YUCK ;) but I guess we can't really avoid such edge cases

Saif807380:
I'll update the test

tests/test_table_name.py
davidism: To be fair, the original name was yuck too.

PR #955, opened 2021-04-14
fix: session get bind signature

### LINKED ISSUE #953: Possible issue on session get_bind
This may be a problem on SQLAlchemy itself, simple example:

``` python
from flask import Flask

app = Flask(__name__)
app.config["SQLALCHEMY_DATABASE_URI"] = "sqlite:///example.sqlite"
db = SQLAlchemy(app)

db.session.get_bind()
```
The output from the execution above:
```
  File "exp.py", line 8, in <module>
    db.session.get_bind()
  File "<string>", line 2, in get_bind
TypeError: get_bind() takes from 1 to 3 positional arguments but 6 were given
```

Should return a SQLAlchemy engine

Environment:

- Python version: 3.7
- SQLAlchemy version: 1.4.7

dpgaspar:
The `SignallingSession` `get_bind` is being called without any parameters, this is probably due to the new SQLAlchemy proxied mechanism for registering scoped sessions 

```
(None, None, None, None, False) {}
```

This is a simple fix that just uses the exact method signature from SQLAlchemy for get_bind of the current session

- fixes #953

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -24,6 +24,14 @@ Unreleased
     ``db.session.commit()`` directly instead. :issue:`216`
 
 
+Version 2.5.2
+-------------
+
+Unreleased
+
+-   Fix session ``get_bind`` signature for SQLAlchemy 1.4. :issue:`953`
+
+
 Version 2.5.1
 -------------
 
--- a/src/flask_sqlalchemy/__init__.py
+++ b/src/flask_sqlalchemy/__init__.py
@@ -172,7 +172,14 @@ class SignallingSession(SessionBase):
             **options,
         )
 
-    def get_bind(self, mapper=None, **kwargs):
+    def get_bind(
+        self,
+        mapper=None,
+        clause=None,
+        bind=None,
+        _sa_skip_events=None,
+        _sa_skip_for_implicit_returning=False,
+    ):
         """Return the engine or connection for a given model or
         table, using the ``__bind_key__`` if it is set.
         """
@@ -191,7 +198,13 @@ class SignallingSession(SessionBase):
                 state = get_state(self.app)
                 return state.db.get_engine(self.app, bind=bind_key)
 
-        return super().get_bind(mapper, **kwargs)
+        return super().get_bind(
+            mapper=mapper,
+            clause=clause,
+            bind=bind,
+            _sa_skip_events=_sa_skip_events,
+            _sa_skip_for_implicit_returning=_sa_skip_for_implicit_returning,
+        )
 
 
 class _SessionSignalEvents:
--- a/src/flask_sqlalchemy/__init__.py
+++ b/src/flask_sqlalchemy/__init__.py
@@ -175,10 +175,8 @@ class SignallingSession(SessionBase):
     def get_bind(
         self,
         mapper=None,
-        clause=None,
-        bind=None,
-        _sa_skip_events=None,
-        _sa_skip_for_implicit_returning=False,
+        *args,
+        **kwargs,
     ):
         """Return the engine or connection for a given model or
         table, using the ``__bind_key__`` if it is set.
@@ -199,11 +197,9 @@ class SignallingSession(SessionBase):
                 return state.db.get_engine(self.app, bind=bind_key)
 
         return super().get_bind(
-            mapper=mapper,
-            clause=clause,
-            bind=bind,
-            _sa_skip_events=_sa_skip_events,
-            _sa_skip_for_implicit_returning=_sa_skip_for_implicit_returning,
+            mapper,
+            *args,
+            **kwargs,
         )
 
 
--- a/tests/test_sessions.py
+++ b/tests/test_sessions.py
@@ -1,6 +1,7 @@
 import random
 
 import sqlalchemy as sa
+from sqlalchemy.engine import Engine
 from sqlalchemy.orm import sessionmaker
 
 from flask_sqlalchemy import SQLAlchemy
@@ -59,3 +60,8 @@ def test_insert_update_delete(db):
 
 def test_listen_to_session_event(db):
     sa.event.listen(db.session, "after_commit", lambda session: None)
+
+
+def test_session_get_bind(app, db):
+    with app.test_request_context():
+        assert isinstance(db.session.get_bind(), Engine)


CHANGES.rst
```diff
@@ -3,6 +3,7 @@ Version 3.0.0
 
 Unreleased
 
+-   Fix session get_bind signature for SQLAlchemy 1.4
```
davidism, line 6: New items go at bottom of changelog.

src/flask_sqlalchemy/__init__.py
```diff
@@ -172,7 +172,14 @@ def __init__(self, db, autocommit=False, autoflush=True, **options):
...
         )
 
-    def get_bind(self, mapper=None, **kwargs):
+    def get_bind(
+        self,
+        mapper=None,
+        clause=None,
+        bind=None,
+        _sa_skip_events=None,
+        _sa_skip_for_implicit_returning=False,
```
davidism, line 181: Why is `**kwargs` not sufficient here? Also, if we're going to capture these, why not pass them on in `super().get_bind()`?

CHANGES.rst
dpgaspar: done

CHANGES.rst
```diff
@@ -23,6 +22,7 @@ Unreleased
 -   Deprecate ``SQLALCHEMY_COMMIT_ON_TEARDOWN`` as it can cause various
     design issues that are difficult to debug. Call
     ``db.session.commit()`` directly instead. :issue:`216`
+-   Fix session get_bind signature for SQLAlchemy 1.4. :issue:`953`
```
davidism, line 25: Use Sphinx syntax:

```
``get_bind``
```

src/flask_sqlalchemy/__init__.py
dpgaspar: That's a very good question, why `**kwargs` is not sufficient.

Basically I found that:

``` python
from flask import Flask
from flask_sqlalchemy import SQLAlchemy, SignallingSession
from sqlalchemy.orm.session import Session
from sqlalchemy.orm import sessionmaker


class ExpSession(SignallingSession):
    def get_bind(self, *args, **kwargs):
        print(args, kwargs)


class ExpSQLAlchemy(SQLAlchemy):
    def create_session(self, options):
        return sessionmaker(class_=ExpSession, db=self, **options)


app = Flask(__name__)
app.config["SQLALCHEMY_DATABASE_URI"] = "sqlite:///example.sqlite"
db = ExpSQLAlchemy(app)

session = db.session

print(session.get_bind())
```

The output is:
```
(None, None, None, None, False) {}
```

CHANGES.rst
davidism: Actually, this can be a new section 2.5.2, it's just a bug fix against a bug fix in 2.5.0.

CHANGES.rst
```diff
@@ -22,7 +22,13 @@ Unreleased
 -   Deprecate ``SQLALCHEMY_COMMIT_ON_TEARDOWN`` as it can cause various
     design issues that are difficult to debug. Call
     ``db.session.commit()`` directly instead. :issue:`216`
--   Fix session get_bind signature for SQLAlchemy 1.4. :issue:`953`
+
```
davidism, line 25: Two spaces between sections.

dpgaspar: done

davidism (dismissed):

davidism:
Needs to be rebased to 2.x branch, since it's a fix against 2.x.

src/flask_sqlalchemy/__init__.py
ashb: @dpgaspar The example you show above seems to behave the same with and without your PR.

ashb, line 181:
```suggestion
    def get_bind(
        self,
        mapper=None,
        *args,
        **kwargs,
```

ashb (changes requested):
Adding `*args` is less fragile

src/flask_sqlalchemy/__init__.py
ashb: And they use `super().get_bind(*args, mapper=mapper, **kwargs)`

ashb: Hmm, I made that work, but now I'm getting "multiple values for mapper"

dpgaspar:
> Needs to be rebased to 2.x branch, since it's a fix against 2.x. You don't need to re-request my review, I get notifications on events already.

I can do that, but if this affects master wouldn't it be desirable to merge this to master, then cherry pick into 2.x?

I can also open a new PR from 2.x to 2.x with this patch?

src/flask_sqlalchemy/__init__.py
ashb: Oh,
 `super().get_bind(mapper, *args, **kwargs)`

ashb: ^^ @dpgaspar

ashb:
I'd love to know _what_ is causing the default args to be put in to args in the first place.

From some playing it's _something_ to do with the scoped_session/session factory.

This doesn't work without this PR or something like it:

```python
session.get_bind()
```

But this does:

```python
session().get_bind()
```

src/flask_sqlalchemy/__init__.py
dpgaspar:
```
TypeError: get_bind() got multiple values for argument 'mapper'
```

ashb: `mapper, *args` vs `*args, mapper` I think (I edited my second comment)

dpgaspar:
> I'd love to know _what_ is causing the default args to be put in to args in the first place.
> 
> From some playing it's _something_ to do with the scoped_session/session factory.
> 
> This doesn't work without this PR or something like it:
> 
> ```python
> session.get_bind()
> ```
> 
> But this does:
> 
> ```python
> session().get_bind()
> ```

For sure, on SQLAlchemy < 1.4 was pretty simple, `get_bind` would get called with:
`return getattr(self.registry(), name)(*args, **kwargs)`

But now, I confess I don't follow what is going on

ashb:
```
[22] > /home/ash/.virtualenvs/flask-sqla/lib/python3.9/site-packages/sqlalchemy/util/langhelpers.py(658)instrument()
-> proxy_fn = _exec_code_in_env(code, env, fn.__name__)
(Pdb++) pp code
'def get_bind(self, mapper=None, clause=None, bind=None, _sa_skip_events=None, _sa_skip_for_implicit_returning=False):\n    return self._proxied.get_bind(mapper, clause, bind, _sa_skip_events, _sa_skip_for_implicit_returning)'
(Pdb++) print(code)
def get_bind(self, mapper=None, clause=None, bind=None, _sa_skip_events=None, _sa_skip_for_implicit_returning=False):
    return self._proxied.get_bind(mapper, clause, bind, _sa_skip_events, _sa_skip_for_implicit_returning)
```

https://github.com/sqlalchemy/sqlalchemy/blob/93ea31df86fc925a9e1df2a7dc80dd39f7ae4b9e/lib/sqlalchemy/util/langhelpers.py#L613

This seems to be the source of the extra arguments.

@zzzeek Is there a particular reason the proxy is including the defaults as explicit parameters, rather than letting python defaults do their thing?

dpgaspar:
@davidism 
I opened another PR from branch 2.x to branch 2.x: https://github.com/pallets/flask-sqlalchemy/pull/956
(it if helps)

ashb (changes requested):
We're going to have to use the *args approach as SQLA 1.3 only has 2 args.

(That, or update the dep on SQLAlchemy to >=1.4)

zzzeek:
> ```
> [22] > /home/ash/.virtualenvs/flask-sqla/lib/python3.9/site-packages/sqlalchemy/util/langhelpers.py(658)instrument()
> -> proxy_fn = _exec_code_in_env(code, env, fn.__name__)
> (Pdb++) pp code
> 'def get_bind(self, mapper=None, clause=None, bind=None, _sa_skip_events=None, _sa_skip_for_implicit_returning=False):\n    return self._proxied.get_bind(mapper, clause, bind, _sa_skip_events, _sa_skip_for_implicit_returning)'
> (Pdb++) print(code)
> def get_bind(self, mapper=None, clause=None, bind=None, _sa_skip_events=None, _sa_skip_for_implicit_returning=False):
>     return self._proxied.get_bind(mapper, clause, bind, _sa_skip_events, _sa_skip_for_implicit_returning)
> ```
> 
> https://github.com/sqlalchemy/sqlalchemy/blob/93ea31df86fc925a9e1df2a7dc80dd39f7ae4b9e/lib/sqlalchemy/util/langhelpers.py#L613
> 
> This seems to be the source of the extra arguments.
> 
> @zzzeek Is there a particular reason the proxy is including the defaults as explicit parameters, rather than letting python defaults do their thing? (i.e. using `*args, **kwargs`)?

I would assume so they get included in the docstrings.     

im not seeing the whole context here, it looks like you're subclassing Session which is fine.  Is it going wrong with scoped_session()?

ashb:
@zzzeek Yeah, subclassing Session.

Given the dynamic function is via a string-eval, that won't show up in docstrings I don't think? Isn't `functools.update_wrapper` or similar a better approach for getting docstrings?

Very rough reproduction steps

```python
class ExpSQLAlchemy(SQLAlchemy):
    def create_session(self, options):
        return sessionmaker(class_=SignallingSession, db=self, **options)


app = Flask(__name__)
app.config["SQLALCHEMY_DATABASE_URI"] = "sqlite:///example.sqlite"
db = ExpSQLAlchemy(app)

session = db.session
# This is fine
print(db.session().get_bind())
# This errors with "wrong number of positional arguments)
print(session.get_bind())
```

Because of the subclassing as it _was_ defined;

```python
     def get_bind(self, mapper=None, **kwargs):
```

that used to work, but on SQLA 1.4 that now doesn't, as the proxy code passes them all as positional.

zzzeek:
not sure if you noticed check out https://github.com/sqlalchemy/sqlalchemy/issues/6285 but I just put out 1.4.8, so might not get to this for some days

dpgaspar:
@zzzeek Thank you

dpgaspar:
Going to close this, since SQLAlchemy is fixed and flask-sqlalchemy master looks good now

PR #956, opened 2021-04-15
fix: session get bind signature

### LINKED ISSUE #953: Possible issue on session get_bind

dpgaspar:
The `SignallingSession` `get_bind` is being called without any parameters, this is probably due to the new SQLAlchemy proxied mechanism for registering scoped sessions 

```
(None, None, None, None, False) {}
```

This is a simple fix that just uses the exact method signature from SQLAlchemy for get_bind of the current session

- fixes #953

Checklist:

### DIFF
--- a/flask_sqlalchemy/__init__.py
+++ b/flask_sqlalchemy/__init__.py
@@ -184,7 +184,12 @@ class SignallingSession(SessionBase):
             bind=bind, binds=binds, **options
         )
 
-    def get_bind(self, mapper=None, clause=None):
+    def get_bind(
+        self,
+        mapper=None,
+        *args,
+        **kwargs,
+    ):
         """Return the engine or connection for a given model or
         table, using the ``__bind_key__`` if it is set.
         """
@@ -202,7 +207,11 @@ class SignallingSession(SessionBase):
             if bind_key is not None:
                 state = get_state(self.app)
                 return state.db.get_engine(self.app, bind=bind_key)
-        return SessionBase.get_bind(self, mapper, clause)
+        return SessionBase.get_bind(
+            mapper,
+            *args,
+            **kwargs,
+        )
 
 
 class _SessionSignalEvents(object):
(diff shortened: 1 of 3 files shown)


flask_sqlalchemy/__init__.py
```diff
@@ -202,7 +209,13 @@ def get_bind(self, mapper=None, clause=None):
             if bind_key is not None:
                 state = get_state(self.app)
                 return state.db.get_engine(self.app, bind=bind_key)
-        return SessionBase.get_bind(self, mapper, clause)
+        return super().get_bind(
```
ThiefMaster, line 212: 2.x still supports python 2, so the super shorthand won't work

dpgaspar: got it!

ashb (changes requested):
We're going to have to use the `*args` approach mentioned https://github.com/pallets/flask-sqlalchemy/pull/955/files#r614057569 as SQLA 1.3 only has 2 args.

(That, or update the dep on SQLAlchemy to >=1.4)

Edit: Daniel gave me access to his fork, and I've made this change there

dpgaspar:
Going to close this, since SQLAlchemy is fixed and flask-sqlalchemy master looks good now. Decided to wait for 2.X

PR #957, opened 2021-04-16
Update __init__.py

SixK:
Fix CoburnJoe fix for NullPool and apply same fix for StaticPool too
- fixes #804


davidism:
I don't understand what this is fixing. The issue you've referred to does not seem related. Be sure to follow the pull request template when opening a pull request, including describing in detail what the fix is.

davidism:
This appears to be https://github.com/CoburnJoe/flask-sqlalchemy/commit/3db800eb5f6dea0f28abc6394e424974914d3e7a, but sets you as the author instead of Joe. Given that he was writing this and indicated there was more he needed to consider, I don't think this PR should be merged.

PR #982, opened 2021-06-04
add mypy to the build and some typing

fuhrysteve:
Since we're dropping support for old versions of python (pre-3.6) I thought it might be appropriate to include some typing in the 3.x series, along with mypy in the build to check typing where appropriate.


davidism:
Thanks for working on this. However, after my experience adding typing to the Pallets libraries, I am only confident in myself adding typing in one go, rather than others adding it bit by bit. Nothing against your work here, just don't feel comfortable merging it.

PR #985, opened 2021-07-21
Update models.rst

arihant2math:
Add part about how to insert records when there are one to many models in the db.


davidism:
Thanks, but I don't want to add more "how to use sqlalchemy" docs, that's already covered in sqlalchemy's docs.

PR #1003, opened 2021-09-23
fix: MetaMixin* accept kwargs

### LINKED ISSUE #1002: Metaclasses do not accept additional keyword arguments
Currently, the metaclass mixins provided by flask-sqlalchemy do not accept additional keyword arguments. Due to that, it breaks compatibility to the `__init_subclass_` hook.

The bug was present in SQLAlchemy, and was fixed in [#5357](https://github.com/sqlalchemy/sqlalchemy/issues/5357), refer to [documentation](https://docs.python.org/3/reference/datamodel.html#creating-the-class-object) about the canonical inclusion of `kwargs`.

~~~python
from sqlalchemy import Column, Integer

db = SQLAlchemy()


class Mixin:
    def __init_subclass__(cls, *, default=None, **kwargs):
        super().__init_subclass__(**kwargs)

        cls.default = default


class Model(db.Model, Mixin, default=2):
    id = Column(Integer, primary_key=True)
~~~

The execution of the snippet will result in:

~~~
  File "[omitted]", line 14, in <module>
    class Model(db.Model, Mixin, default=2):
TypeError: __init__() got an unexpected keyword argument 'default'
~~~

The instantiation of the class should not fail, and `default` should be propagated to the `__init_subclass__` hook.

Environment:

- Python version: 3.9.7
- SQLAlchemy version: 1.4.23

indietyp:
The Metaclass Mixin classes now accept and propagate additional keywords provided.

The name `**kw` has been chosen to stay consistent with [SQLAlchemy](https://github.com/sqlalchemy/sqlalchemy/blob/master/lib/sqlalchemy/orm/decl_api.py#L56).

- fixes #1002

Checklist:

- [ ] Run `pytest` and `tox`, no tests failed.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,7 @@ Unreleased
 -   Deprecate ``SQLALCHEMY_COMMIT_ON_TEARDOWN`` as it can cause various
     design issues that are difficult to debug. Call
     ``db.session.commit()`` directly instead. :issue:`216`
+-   Support ``__init_subclass__()`` hooks in models :issue:`1002`
 
 
 Version 2.5.1
--- a/src/flask_sqlalchemy/model.py
+++ b/src/flask_sqlalchemy/model.py
@@ -48,11 +48,11 @@ def camel_to_snake_case(name):
 
 
 class NameMetaMixin(type):
-    def __init__(cls, name, bases, d):
+    def __init__(cls, name, bases, d, **kw):
         if should_set_tablename(cls):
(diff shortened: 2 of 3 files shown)


indietyp:
The current problem is that tox fails on old versions of SQLAlchemy, where the fix wasn't introduced yet, but that is more an issue with SQLAlchemy, not with the implementation.

davidism:
Fixed in #1087

PR #1076, opened 2022-08-11
Update docs dependencies

sirosen:
Currently doc builds appear broken. I think this is producing a poor signal:noise on this repo and probably making it harder for maintainers to handle PRs.

`pip-compile -U requirements/docs.in` to get latest.
Importantly, this picks up pallets-sphinx-themes==2.0.2 so that doc builds now work with the latest jinja2 versions.

### DIFF
--- a/requirements/docs.txt
+++ b/requirements/docs.txt
@@ -6,45 +6,45 @@
 #
 alabaster==0.7.12
     # via sphinx
-babel==2.9.1
+babel==2.10.3
     # via sphinx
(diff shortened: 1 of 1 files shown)


sirosen:
I just realized that #1019 has been sitting, so possibly this should be closed in favor of several of the dependabot PRs. I'll leave this here though, in case this is more convenient.

davidism:
Just did this manually for everything.

PR #1087, opened 2022-09-18
implement metadata per bind, refactor entire extension

### LINKED ISSUE #70: Support __iter__ on Pagination instances.
I never remember to call .items, and I was just reminded of this when getting a new developer spun up.  Instead of:

```
for obj in MyModel.query.paginate(1).items: pass
```

I expect:

```
for obj in MyModel.query.paginate(1): pass
```

... because SQLAlchemy's query API does this everywhere.  This fails, because a Pagination instance is not iterable.

### LINKED ISSUE #327: Specify custom session class
Hello,

SQLAlchemy allows us to use custom session classes in case we wanna override default behavior:

``` python
sqlalchemy.orm.scoped_session(
    sqlalchemy.orm.sessionmaker(
        autocommit=False,
        autoflush=False,
        bind=self.engine,
        class_=NoDeleteSession  # Custom session class
    )
)
```

This is critical, as in most business programs operations like `DELETE` do not physically delete data, but mark it as deleted instead:

``` python
class NoDeleteSession(sqlalchemy.orm.session.Session):
    def __init__(self, bind=None, autoflush=True, expire_on_commit=True,
                 _enable_transaction_accounting=True,
                 autocommit=False, twophase=False,
                 weak_identity_map=True, binds=None, extension=None,
                 info=None, query_cls=sqlalchemy.orm.query.Query):

        self._persist_deleted = set()

        super(NoDeleteSession, self).__init__(
            bind=bind,
            autoflush=autoflush,
            expire_on_commit=expire_on_commit,
            _enable_transaction_accounting=_enable_transaction_accounting,
            autocommit=autocommit,
            twophase=twophase,
            weak_identity_map=weak_identity_map,
            binds=binds,
            extension=extension,
            info=info,
            query_cls=query_cls
        )

    def delete(self, instance):
        # Do not actually delete the instance. Set the is_deleted flag instead.
        instance.is_deleted = True
        self._persist_deleted.add(instance)
```

Going through the flask-sqlalchemy source code and documentation, it seems that this isn't currently supported. Can we have this functionality?

Thanks.

### LINKED ISSUE #417: modifications to backref made inside _wrap_with_default_query_class(fn, cls) get lost
```
def _wrap_with_default_query_class(fn, cls):
    @functools.wraps(fn)
    def newfn(*args, **kwargs):
        _set_default_query_class(kwargs, cls)
        if "backref" in kwargs:
            backref = kwargs['backref']
            if isinstance(backref, string_types):
                backref = (backref, {})
            _set_default_query_class(backref[1], cls)
        return fn(*args, **kwargs)
return newfn
```

In the code above `backref` is extracted from `kwargs`, modified if it's a string type, but the modifications are never reintroduced to `kwargs`.

### LINKED ISSUE #508: session.remove called too early when there are stacked application contexts
Currently flask-sqlalchemy will call `session.remove()` when `app.teardown_appcontext` is called. Because application contexts can be stacked, it means that we will remove the session when the inner most application context is popped, while there still might be other active contexts.

Aren't sessions supposed to be scoped per application context?

Here's an example to show the issue:

```python
from flask import Flask

app = Flask(__name__)
db = SQLAlchemy()
db.init_app(app)


class User(db.Model):
    name = db.Column(db.String())


def test_session_remove_example():
    with app.app_context():
        db.create_all()

        user = User(name='a')
        db.session.add(user)
        assert user in db.session

        with app.app_context():
            assert user in db.session

        # This will fail, because the session was removed.
        assert user in db.session
```

### LINKED ISSUE #622: Very poor performance of Pagination.{pages,iter_pages}
The performance of those two methods is really really bad when you have thousands of pages. In my case, with 100k pages, painting the paginator takes over 300 ms.

`Pagination.iter_pages` is slow by itself. `Pagination.pages` is not that slow but it's called once for every page from `Pagination.iter_pages` when you consume the iterator, and calling it 100k times really adds up. Python is a very slow language and calling the same method so many times is undesirable even if [the method seems simple](https://github.com/mitsuhiko/flask-sqlalchemy/blob/381653489af8ee8aa26340a7c3fe0a0dced64258/flask_sqlalchemy/__init__.py#L325-L332).

The low-hanging fruit here is caching the result of `Pagination.pages`. That fixes more than half of the problem (only the call to `Pagination.iter_pages` remains). I did so with this subclass, which might be helpful to someone. (I'm far from an expert in Python, so this may hurt your eyes)

```python
class MyPagination(Pagination):

    def __init__(self, query, page, per_page, total, items):
        super().__init__(query, page, per_page, total, items)
        self._pages = super().pages

    @property
    def pages(self):
        return self._pages
```

But of course a better solution is desirable.

### LINKED ISSUE #941: Our implementation of binds can cause table name conflicts
This is a description of the issue that #222 tries to solve. After investigating further, and based on differences between SQLAlchemy 1.3 and 1.4, I don't think I'll be able to merge that PR so I'm writing up more here.

We have the `SQLALCHEMY_BINDS` config mapping keys to engine URLs. Each model can have an optional `__bind_key__` attribute. We override `Session.get_bind` to look for this key and choose the engine from the config. In this way, you can define models that are present in different databases.

This is slightly different than SQLAlchemy itself. There the `Session(binds={})` maps classes to engines. So a specific model could have a specific engine, or a base class could be used to map all its models to the same engine. The configuration is done when setting up the session and engines, not when defining the models and tables.

Our way causes issues when two models with the same name / table name are defined that belong to separate binds. The `Model` base class has one `metadata` associated with it, and all names registered with a metadata must be unique. It also makes it possible to write foreign keys between models that will end up using separate engines, which won't work. When using plain SQLAlchemy, you would create separate declarative bases and bind each of them to a different engine, but in Flask-SQLAlchemy there is only one `db.Model` base class.

#222 addresses this by creating a different metadata in the metaclass when creating a model with a different bind key. I wasn't super comfortable with that, but the release of SQLAlchemy 1.4 made it more clear why. In 1.4, it uses a new `registry` object, and `Base.metadata` is essentially an alias to `Base.registry.metadata`. Looking back at how 1.3 did it, `Base._decl_class_registry` was the equivalent, and it wasn't being overridden by #222, so you'd still have names overwriting each other in the registry if not the metadata. With 1.4, we'd  need to override `registry`, not `metadata` directly, and looking at #222's current implementation this seems even more complex and messy. And if we want to support SQLAlchemy <= 1.3 we need to detect and override both the old and new implementations.

The problem is that our `__bind_key__` is only available after class creation has started, but `metadata`/`registry` was created before that when creating the declarative base. There's a disconnect between when we have the information to know what to create, and when it needs to be created. *Maybe* it could be addressed with metaclass trickery to substitute a different base class when creating a subclass with a different key, but I haven't investigated very far and I'm not particularly enthusiastic about trying to deal with that complexity.

Alternatively, we could make `db.make_declarative_base` more of a public API (or a new method) and have it take a `bind_key` parameter. So when you want to use a separate bind, instead of inheriting `db.Model`, inherit `db.get_base_model("key")`. We'd probably want to disallow setting `__bind_key__` manually and show a message saying what to do instead, except there are also valid reasons to have models in the same metadata use different binds, as long as there's no conflicts. However, this seems pretty confusing to teach users, I foresee it causing a bunch of new questions even as it solves the current problem.

### LINKED ISSUE #944: scoped_session’s scopefunc uses LocalStack() instead of AppContext()
Whenever multiple app contexts are pushed to the context stack the session handling fails. Although this is not a likely use-case outside of tests it is super-hard to debug *when* it happens.

The code for handling the session life-time seems to assume that it should remove a session whenever an app context is teared down. With `connection_stack.__ident_func__` as scopefunc for the scoped_session we don’t get one session per app context but one session per thread. That is because the function references the stack itself and not the app context on the stack.

This becomes relevant when more than one app-context is pushed to the stack. In such cases the session is removed when the innermost app context is destroyed. This then can lead to various different errors in the code relying on the outer app context’s session.

This is especially hard to debug, because everything in Flask-SQLAlchemy looks like the session is handled per app-context while this is actually not the case.

-----

For context here are the two relevant comments of #379 . Since the issue there is closed I can’t post there so I’m opening this new issue.

> also I dont get the design behind the choice of scopefunc, by default scopefunc is `connection_stack.__ident_func__` which is effectively implemented with thread.get_ident, means that app would get same session within same thread, no matter how many app context on the stack, I think `scopefunc = lambda: id(flask_sqlalchemy.connection_stack.top)` is more desirable.
— @georgexsh

and the answer to this was was

> I'm going to need an example of how to run Flask in such a way that the same app context is reused for multiple requests (not for multiple manual request contexts within a single app context or test, but for actual requests). I can't reproduce this error and using the app context should be correct. The app context is for global app state during an interaction (such as a request or cli command).
— @davidism

### LINKED ISSUE #967: Misleading __repr__
The standard repr is misleading such that:

```python
instance_state = inspect(item)
print(f"{item = }; {instance_state.transient = }; {instance_state.pending = }")  
# item = <Competition (transient 140147915565184)>; instance_state.transient = False; instance_state.pending = True
```

I spent a small amount of time debugging why my instance was transient after I thought it should already be in the session and therefore `pending`. SQLAlchemy defines `transient` as:

> an instance that’s not in a session, and is not saved to the database; i.e. it has no database identity. The only relationship such an object has to the ORM is that its class has a Mapper associated with it.

On the surface, the repr looks as though it is presenting the [instance states](https://docs.sqlalchemy.org/en/14/orm/session_state_management.html#quickie-intro-to-object-states) as they are defined by SQLAlchemy. Given that the word "transient" has a very specific meaning, overloading it here with a different meaning doesn't quite feel right to me.

Can we afford to be more specific about object state in the repr or use an alternate word to "transient"? Or has the ship sailed?

https://github.com/pallets/flask-sqlalchemy/blob/9a467c19a6b58bdbbff55eb16dcfb6337a51ace6/src/flask_sqlalchemy/model.py#L132-L140

davidism:
This began as an attempt to fix the longstanding metadata per bind issue. Based on the pervasiveness of that change throughout the code, and on observations of the organization and maintainability of the code, I ended up refactoring the entire project. I've also rewritten all the tests and documentation. This PR addresses or invalidates the majority of current open issues and PRs. Below is a summary and reasoning for _most_ of the changes, but check the changelog as well.

### Compatibility

This will be Flask-SQLAlchemy 3.0. While the basic API (config, `db.Model`, `db.session`, etc.) remains the same, the rest of the code has changed significantly. I have tried to add in deprecation warnings where it was feasible. I will be making an alpha release based on this PR for users to try out and report any major incompatibilities that cannot be adapted to.

Many things have been moved to submodules and renamed. The signatures of some methods have changed, such as switching to keyword-only arguments. All setup methods that were used to create objects have been renamed with leading underscore, and are no longer considered public API. Therefore, projects that were subclassing or calling those methods to do advanced customization will encounter incompatibilities. Customization of all the objects managed by SQLAlchemy should be possible through arguments to `__init__`.

### Static Typing

I added static type annotations to the entire library. For the most part, `mypy --strict src` passes. However, both mypy and pyright have behaviors that will frustrate users if I export the types. For example, mypy will not allow subclassing `db.Model`. Therefore, I am not enabling the annotations for other projects for now. SQLAlchemy 2 is adding their own static types, so I'd be happy to get help with this from the community.

### Minimum Versions

* Python minimum version is 3.7. Python 2, 3.4, 3.5, and 3.6 is dropped.
* Flask minimum version is 2.2 to take advantage of new app context tracking and async contexts.
* SQLAlchemy minimum version is 1.4. 1.4 begins to change in preparation for the 2.0 release. Try running your project in [SQLAlchemy 2 preview mode](https://docs.sqlalchemy.org/en/14/changelog/migration_20.html).

### App Context

The extension no longer stores the app object if it is initialized as `SQLAlchemy(app)`. An active app context is _always_ required when using the session or engines. The session is scoped to the application context, rather than the current thread. By requiring an app context, this ensures that the session is always cleaned up after it is used, whether in a test, request, or CLI command.

Besides the correctness issue, I think a large amount of the confusion over when the app context was needed was because there were two different modes, and users might see an example assuming one mode while some other tutorial they followed used the other. Now, there is only one way to use it: with the context.

On a related note, Flask 2.2 changed how app contexts are cleaned up in debug mode. The old way may have prevented returning connections to the pool in debug mode, it will be interesting to see if users have fewer connection issues.

### Metadata per Bind

One of the big issues in Flask-SQLAlchemy was that one metadata was used for all model and tables in all binds. This meant that tables in different engines could not have the same name, and caused other issues where things were expected to be more separated.

The extension now has a `db.metadatas` dict mapping bind keys to metadata instances. `db.metadata` is the default metadata like `db.engine`. When a model or table uses a bind key, it will create a new metadata object if it doesn't exist yet. The default metadata can still be customized with the `metadata` argument. Other metadatas will copy the default metadata's `naming_convention`.

`db.engines` is a dict mapping engines to bind keys, like `db.metadatas`. `get_engine` is deprecated, engines and metadatas should be looked up as the keys in the dicts instead if needed. There is no longer a separate "state" object with "connectors", the extension is stored directly in `app.extensions["sqlalchemy"]` and the engines are stored in a weakref dict for each app on the extension.

Engines are created during `init_app` instead of on first access. This is safe

This makes the implementation of `create_all`, `delete_all`, and `reflect` much simpler, as it can now just call the appropriate method on each metadata and engine pair, rather than on the same metadata with different engines. It's possible to inspect the tables for a bind by looking at `db.metadatas[key].tables`.

The `bind_key` is stored on each `metadata.info`. The session always looks up the key on the metadata, not on the model or table.

fixes #941 

### Pagination

* `Pagination` is iterable. fixes #70
* `iter_pages` is more efficient. fixes #622
* `Pagination.apply_to_query` classmethod can be used instead of `paginate.query`.
* There is now documentation for using pagination.

### Query

* Added `query.one_or_404` to match the other 404 methods.
* `Query` is renamed from `BaseQuery`.
* `db.relationship` sets the query class on backrefs as well. fixes #417

Note that SQLAlchemy 2 considers the query interface legacy. They won't continue to develop it and may deprecate it in the future. They are moving to using `session.execute(select())` instead. I don't plan to deprecate `db.Query`, `db.session.query`, `db.relationship query_class`, or `Model.query` until they do, but I will be making a separate PR to implement the 404 and paginate methods for `select()` objects.

### Session

* The session is scoped to the current application context instead of the thread. Therefore accessing the session requires an active application context. fixes #508 fixes #944
* `SignallingSession` is renamed to `Session`
* `get_bind` more closely matches the base implementation.
* The session class can be customzied by passing `session_options={"class_": SessionSubclass}`. fixes #327

### Configuration

Init arguments are used to customize objects managed by the extension, which are common to all apps initialized on the extension. App config is used to customize the engines, which are specific to each app. Configuration is only read during init, it cannot be changed after init.

* `engine_options` argument sets defaults for all engines rather than overriding config.
* `SQLALCHEMY_BINDS` values can be dicts (instead of only connection strings) to set the engine options for a specific bind. In that case, it must have the `"url"` key in the options.
* `SQLALCHEMY_ENGINE_OPTIONS` only applies to the default bind, and overrides anything set by `SQLALCHEMY_BINDS[None]`.
* Connection strings can be strings or SQLAlchemy `URL` objects.
* `SQLALCHEMY_DATABASE_URI` overrides a `"url"` in `SQLALCHEMY_ENGINE_OPTIONS`.
* `SQLALCHEMY_ECHO` sets `pool_echo` in addition to the engine's `echo`.

### Other Changes

* `get_debug_queries` is renamed to `get_recorded_queries` to better describe what it returns. It now returns a dataclass instead of a namedtuple. The `context` attribute is renamed to `location`. Calculating `location` is more accurate for apps defined in subpackages.
* Names from `sqlalchemy` and `sqlalchemy.orm` are aliased using `__getattr__` rather than setting tons of attributes on the extension object.
* Model `__repr__` distinguishes between transient and pending objects. fixes #967
* `db.Table` is a subclass instead of a function.
* The `bind` argument to some methods was renamed to `bind_key` to distinguish between the Flask-SQLAlchemy feature and SQLAlchemy's binds.
* Rewrote all the tests to use modern pytest patterns.
* Rewrote all the docs. Focused on the use of Flask-SQLAlchemy itself, rather than trying to be a SQLAlchemy tutorial.

### Further Considerations

I'll make some more PRs for some issues this didn't address, since they could be made outside the main refactor. Additionally, I had the following notes on potential further changes. I'll make issues for these, just wanted to introduce them since it's what I noticed during this refactor.

* To support SQLAlchemy 2, we need to support pagination and 404 methods on `select`. After talking with SQLAlchemy maintainers, I think I'll add them as `db.get_or_404`, etc. since adding these methods to the session or result objects isn't intended.
* A pattern I use a lot is to write a little `app.shell_context_processor` to add `db` and all the model classes to the `flask shell`.
* Add a CLI command to call `create_all`. Should we also have `drop_all` with a confirmation?
* It would be possible to add some global proxies `current_db` and `db_session`.
* Should we deprecate everything related `Query` since it's legacy? Not yet, but I'll watch what SQLAlchemy does.
* We could make it easier to enable SQLAlchemy 2.0 preview mode, but I'm not sure it's worth it. Currently it can be done with `engine_options={"future": True}, session_options={"future": True}`.
* Pagination: Does it make sense for `per_pages` to allow 0? Does it make sense for `max_per_page` to have a default, like 100? Should `pages` be `None` not 0 if `total` is `None` or 0?
* Stop enabling `SQLALCHEMY_RECORD_QUERIES` by default in debug and testing mode. It only makes sense to enable it and incur the performance hit if you're using Flask-DebugToolbar or otherwise inspecting the info.


jeffwidman:
Thanks for all the hard work here @davidism!

davidism:
A prerelease with these changes, version 3.0.0a1, is now available on PyPI. Please use the prerelease to test your project ahead of the final release.

```
pip install -U --pre flask-sqlalchemy
```

PR #1107, opened 2022-09-27
use pyproject.toml and PDM

davidism:
Use the `pyproject.toml` standard to specify project metadata and dependencies. Use [PDM](https://pdm.fming.dev/latest/) to lock development dependencies and manage the environment.

Tool configuration was moved to `pyproject.toml`, except for flake8, which does not support that file. Flake8 config remains in `setup.cfg`. `setup.py` remains for GitHub's dependency graph, although PDM might write a GitHub action for updating the [dependency graph](https://github.com/github/dependency-submission-toolkit#writing-your-own-dependency-submission-action) eventually.

Read the Docs doesn't support non-pip tools natively, but they recently added support for custom build steps. This uses `post_install` to install PDM and run `pdm sync -dG docs` after RTD sets up the virtualenv and installs its own packages.

Contributors can use PDM to set up a virtualenv and sync development dependencies. [Install PDM](https://pdm.fming.dev/latest/#installation), then do `pdm sync`. This will create a virtualenv in the `.venv` folder if it doesn't exist, then install the locked versions of all development dependencies, and install the project in editable mode. Commands are run with `pdm run ...` without requiring activating the virtualenv. Alternatively, you can use a tool like `pew` to manage virtualenvs, PDM will detect an active virtualenv and use that instead, and `pip install -e .` will continue to work as well.

Common commands:

* `pdm sync` install the current set of locked dependencies
* `pdm run pytest`
* `pdm run mypy`
* `pdm run tox -p` run all test envs in parallel
* `cd docs && make html` make the docs locally, caching changes

Advanced / maintainer commands:

* `pdm update --update-all` update all versions in the lock file
* `eval $(pdm venv activate)` activate the virtualenv so `pdm run ...` isn't needed
* `pew workon flask-sqlalchemy` use pew to manage the virtualenv and switch to the project, then `pdm sync` will install to that env.


davidism:
Looks like Read the Docs is going to be more difficult. It doesn't activate the virtualenv it creates, so PDM creates its own env.

frostming:
You may be interested: https://github.com/pdm-project/pdm/discussions/1365#discussioncomment-3581356

davidism:
Thanks, docs are building now. Reported this to RTD here: https://github.com/readthedocs/readthedocs.org/issues/9629

PR #1116, opened 2022-10-06
Quickstart: switch .first() to .scalar() for querying single record

### LINKED ISSUE #1115: Quickstart doc should instruct to use .scalar() instead of .first() for single item queries
Quickstart doc should instruct to use .scalar() instead of .first() for single item queries.

With new query api .first() returns a Row object, not an instance of whatever model was queried, so it can't simply be passed to templates and used like `Author.name`, `Post.title`.

Environment: N/A, just docs

carc1n0gen:
- fixes #1115

Checklist:

change.


davidism:
Don't need a change entry for this, docs update as soon as they're merged.

carc1n0gen:
Updated

PR #1170, opened 2023-02-26
Rename `first_or_404` and `one_or_404` to `scalar_or_404` and `scalar_one_or_404`

greyli:
The underly implementation of `first_or_404` and `one_or_404` is use `scalar()` and `scalar_one()`. I think it will be more clear to rename them to `scalar_or_404` and `scalar_one_or_404`, so users won't expect or worry `first_or_404` and `one_or_404` will return a Row object/tuple.

It seems reasonable to change the old names to match the new behaviors of SQLAlchemy 2.x style. The old names can be kept for backward compatibility.

I will add the tests, and update the docs if this change will be accepted.

Checklist:

- [ ] Add tests that demonstrate the correct behavior of the change. Tests should fail without the change.

### DIFF
--- a/src/flask_sqlalchemy/extension.py
+++ b/src/flask_sqlalchemy/extension.py
@@ -640,7 +640,7 @@ class SQLAlchemy:
 
         return value
 
-    def first_or_404(
+    def scalar_or_404(
         self, statement: sa.sql.Select[t.Any], *, description: str | None = None
     ) -> t.Any:
         """Like :meth:`Result.scalar() <sqlalchemy.engine.Result.scalar>`, but aborts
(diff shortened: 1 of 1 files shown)


pamelafox:
I chatted with @davidism about this at a conference this weekend. The current first_or_404 and one_or_404 naming is not perfect, but it's sufficient for the most standard usage by developers, and they will get an error in development if they use it incorrectly, so there's not a risk of production errors.

Going forward, now that Flask-SQLAlchemy-Lite exists, we would recommend developers to use that instead, as it doesn't provide these wrapper functions and developers can use the full SQLAlchemy functionality however they like. We can start pointing developers to that from the front page of the FSA docs ("For new projects, we recommend starting with Flask-SQLAlchemy-Lite..."). The new Lite version de-couples our interface much more from the SQLAlchemy interface, so that we won't have to worry about updating names, classes, methods, etc.

greyli:
Sounds good. And thanks for mentioning Flask-SQLAlchemy-Lite, so I could consider using it instead of Flask-SQLAlchemy for the new version of my Flask book.

Issue #1172, opened 2023-03-01
Reload the DB URL/bind config when calling `db.create_all`


greyli:
After 3.0, the engines is created in `init_app`, so there is no way to update the configuration after instantiating the extension object. However, we do need to update the config in some use cases. For example, in a simple application without using the app factory:

```py
app = Flask(__name__)
app.config['SECRET_KEY'] = os.getenv('SECRET_KEY', 'secret string')
app.config['SQLALCHEMY_DATABASE_URI'] = os.getenv(
'DATABASE_URL', 'sqlite:////' + os.path.join(app.root_path, 'data.db')
)
db = SQLAlchemy(app)
```

In the test, you will want to use a different database URL, so you reset the config `SQLALCHEMY_DATABASE_URI`:

```py
from app import app, db

class NotebookTestCase(unittest.TestCase):

    def setUp(self):
        self.context = app.test_request_context()
        self.context.push()
        self.client = app.test_client()
        app.config.update(
            TESTING=True,
            WTF_CSRF_ENABLED=False,
            SQLALCHEMY_DATABASE_URI='sqlite:///:memory:'  # <--
        )
        db.create_all()
```

It works in 2.x, but not 3.x.

I think we need to reload the config (`SQLALCHEMY_DATABASE_URI` and `SQLALCHEMY_BINDS`) before calling `create_all`, but it seems there is no way to update the engine URL after creating it.

davidism:
This is intentional. Factory should create app, set config, then init extensions.

davidism:
There's a few other things that are subtly wrong with your test setup. You call `test_request_context` first, then call `config.update`. But `test_request_context` uses config, what if you wanted to test a different `SERVER_NAME`, for example? You always need to do things in the order create app, set config, run test. Also, you're unconditionally pushing a request context, which means that the app context will never get torn down between tests, which will cause issues.

davidism:
The problem is that Flask-SQLAlchemy allows configuring the entire set of engine options, but once engines are created they're part of SQLAlchemy, not directly configurable. Determining if any engine configuration has changed is not trivial. It would have to be checked on each access to each engine, even though this would be unnecessary overhead for every request _except_ in tests that change engine config.

This is a problem for any extension that manages third-party objects that live longer than the app/request context, not just Flask-SQLAlchemy.

greyli:
[...]

PR #1175, opened 2023-03-02
Add tips for overwriting configuration in tests

greyli:
Related issue: #1172


davidism:
I'm ok with the message, but I'd rather add it to a new page about testing as part of #1171.

greyli:
OK, closing it now. Maybe we won't need this if we can support recreating the engine.

PR #1188, opened 2023-04-04
Document using multiple sessions

danstewart:
Added some documentation for using multiple sessions/transactions in the one request.
Not sure if this is the best way of doing things but seems to be working fine for me.

Checklist:

- [ ] Add tests that demonstrate the correct behavior of the change. Tests should fail without the change.
- [ ] Run `pre-commit` hooks and fix any issues.
- [ ] Run `pytest` and `tox`, no tests failed.

### DIFF
--- a/docs/contexts.rst
+++ b/docs/contexts.rst
@@ -112,4 +112,3 @@ to the new context and disconnected when the context is popped.
 
     # Now commit `user`
     db.session.commit()
-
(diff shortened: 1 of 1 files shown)


davidism:
It's not clear from the description or the docs why this is needed. And as you admit, it's not clear this is the correct way to do so or to solve the issue you're actually having. I'd prefer to leave this out of the docs.

danstewart:
No worries, I'll raise an issue with my use case for some further discussion :+1:

PR #1206, opened 2023-05-06
Fix #1205

### LINKED ISSUE #1205: Little problem with the type hint
image
sqlalchemy does not exporse orm as a clear attr, so the type hint there is not working.
now we change that to this with full path.
image
And it works out with the type hint.

The main problem with that might be the use of some thing like `db.session`, which I use a lot in project to do CRUD operations. Previously we have that. It will stock since sa don't have orm attr. It's a module.

https://github.com/pallets-eco/flask-sqlalchemy/blob/d137db10e733daf1fd2ad24dd59aedc3072d48e5/src/flask_sqlalchemy/extension.py#L325

And I think we should make this little change to provide more type hint info.

Environment:

- Python version: 3.9.16
- Flask-SQLAlchemy version: 3.0.3
- SQLAlchemy version: 2.0.0 and above

Zoupers:
- fixes #1205

Checklist:

- [x] ~~Add tests that demonstrate the correct behavior of the change. Tests should fail without the change.~~
- [x] ~~Add or update relevant docs, in the docs folder and in code.~~
- [x] ~~Add an entry in `CHANGES.rst` summarizing the change and linking to the issue.~~
- [x] ~~Add `.. versionchanged::` entries in any relevant code docs.~~

### DIFF
--- a/src/flask_sqlalchemy/cli.py
+++ b/src/flask_sqlalchemy/cli.py
@@ -5,12 +5,18 @@ import typing as t
 from flask import current_app
 
 
+if t.TYPE_CHECKING:
+    from .extension import SQLAlchemy as SQLAlchemy
+else:
+    SQLAlchemy = t.Any
+
+
 def add_models_to_shell() -> dict[str, t.Any]:
     """Registered with :meth:`~flask.Flask.shell_context_processor` if
     ``add_models_to_shell`` is enabled. Adds the ``db`` instance and all model classes
     to ``flask shell``.
     """
-    db = current_app.extensions["sqlalchemy"]
+    db: SQLAlchemy = current_app.extensions["sqlalchemy"]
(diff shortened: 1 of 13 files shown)


Zoupers:
Delete line means I think there is no need to do that.

davidism:
Thanks, but this has a lot of changes that aren't related to the issue, and I'd rather use `import sqlalchemy.orm as sa_orm`.

davidism:
Also, when opening a pull request, make sure to create a new branch for your work, and give the PR a descriptive title.

Zoupers:
@davidism The changes that not related to the issue, means the test and ?

davidism:
The change to `cli.py`, just leave that as it currently is.

Zoupers:
> The change to `cli.py`, just leave that as it currently is.

OK. I will do it now.

PR #1207, opened 2023-05-06
Fix little problem with type hints and adopt a test to make sure it works in all platform

### LINKED ISSUE #1205: Little problem with the type hint

Zoupers:
- fixes #1205
- fixes a test that cannot work on windows

Checklist:

- [x] ~~Add tests that demonstrate the correct behavior of the change. Tests should fail without the change.~~
- [x] ~~Add or update relevant docs, in the docs folder and in code.~~
- [x] ~~Add an entry in `CHANGES.rst` summarizing the change and linking to the issue.~~
- [x] ~~Add `.. versionchanged::` entries in any relevant code docs.~~

Delete line means I think there is no need to do that.


pamelafox:
@Zoupers Is this ready for review? And can you specify in the description or inline, which test is the one that didn't work on Windows?

src/flask_sqlalchemy/query.py
```diff
@@ -2,16 +2,15 @@
...
+import sqlalchemy.exc as sa_exc
+import sqlalchemy.orm as sa_orm
 from flask import abort
 
 from .pagination import Pagination
 from .pagination import QueryPagination
 
 
-class Query(sa.orm.Query):  # type: ignore[type-arg]
+class Query(sa_orm.Query[t.Any]):
```
pamelafox, line 13: This seems correct per https://github.com/sqlalchemy/sqlalchemy/blob/1c3dbf39ec2d72ea70c0f74fa2e6dc537e76700b/lib/sqlalchemy/orm/query.py#L151

tests/test_record_queries.py
```diff
@@ -24,5 +26,5 @@ class Example(db.Model):
     assert "FROM example" in info.statement
     assert info.parameters[0][0] == 5
     assert info.duration == info.end_time - info.start_time
-    assert "tests/test_record_queries.py:" in info.location
+    assert os.path.join("tests", "test_record_queries.py:") in info.location
```
pamelafox, line 29: I assume this is the test that did not previously work in Windows.

src/flask_sqlalchemy/query.py
Zoupers: According to this, https://github.com/sqlalchemy/sqlalchemy/blob/4808acfe1b88854fce47b792ef9995f1921426e7/lib/sqlalchemy/orm/query.py#L77
which is the version(`1.4.18`) specified in the `requirements/tests-min.in`, it's not correct. I will fix it.

Zoupers:
I have fixed the bug and passed the corresponding tests locally. I apologize for any inconvenience caused.

pamelafox:
@Zoupers Thanks! I think we may upgrade min version of SQLAlchemy to 2.0 in the next version in order to formally support it, but for this minor release, it's better to just get it working for 1.4. This is looking good now.

pamelafox:
I experimented a bit to see if we could automatically test that this works correctly, but we're not currently running pyright on the code, and if we did, we'd have a bunch of other issues that can't be ignored as easily as they are in mypy (as pyright doesn't have the same specific error codes), so I think it's fine to not have a test for the pyright-specific typing right now.

However, if there's a regression in the future, we might consider adding pyright env to tox.

PR #1211, opened 2023-05-31
Fixed clause_to_engine bug not finding the right db bind key

### LINKED ISSUE #1210: Bug with getting the correct bind when clause is a statement (like Insert)
On a Flask app with multiple databases where you have models that have different `__bind_key__` attributes, if you perform a table operation on the model that has its table in a different database `get_bind` cannot find the correct bind key and engine and this raises an OperationError: no such table.

```
@pytest.mark.usefixtures("app_ctx")
def test_session_multiple_dbs(app: Flask) -> None:
    app.config["SQLALCHEMY_DATABASE_URI"] = "sqlite:///"
    app.config["SQLALCHEMY_BINDS"] = {"db1": "sqlite:///"}

    db = SQLAlchemy(app)

    class User(db.Model):
        id = sa.Column(sa.Integer, primary_key=True)
        name = sa.Column(sa.String(50), nullable=False)

    class Product(db.Model):
        __bind_key__ = "db1"
        id = sa.Column(sa.Integer, primary_key=True)
        name = sa.Column(sa.String(50), nullable=False)

    db.create_all()

    db.session.execute(User.__table__.insert(), [{"name": "User1"}, {"name": "User2"}])
    db.session.commit()
    users = db.session.execute(db.select(User)).scalars().all()
    assert len(users) == 2

    db.session.execute(
        Product.__table__.insert(), [{"name": "Product1"}, {"name": "Product2"}]
    )
    db.session.commit()
    products = db.session.execute(db.select(Product)).scalars().all()
    assert len(products) == 2

```

In the following test the first insert on User passes, but the second insert on Product fails with the no such table error. This is because the _clause_to_engine function doesn't support cases where clause has a table attribute that holds a bind_key.

Environment:

- Python version: >= 3.8
- Flask-SQLAlchemy version: >=3.0.0
- SQLAlchemy version: >= 1.4.3

ivan-gerov:
- fixes #1210

Checklist:

- [X] Add or update relevant docs, in the docs folder and in code.
- [  ] Add an entry in `CHANGES.rst` summarizing the change and linking to the issue.
- [X] Add `.. versionchanged::` entries in any relevant code docs.

### DIFF
--- a/src/flask_sqlalchemy/session.py
+++ b/src/flask_sqlalchemy/session.py
@@ -79,14 +79,21 @@ class Session(sa_orm.Session):
 
 
 def _clause_to_engine(
-    clause: t.Any | None, engines: t.Mapping[str | None, sa.engine.Engine]
+    clause: t.Any | None,
+    engines: t.Mapping[str | None, sa.engine.Engine],
 ) -> sa.engine.Engine | None:
     """If the clause is a table, return the engine associated with the table's
     metadata's bind key.
     """
-    if isinstance(clause, sa.Table) and "bind_key" in clause.metadata.info:
-        key = clause.metadata.info["bind_key"]
+    table = None
+    if clause is not None:
+        if isinstance(clause, sa.Table):
+            table = clause
+        elif hasattr(clause, "table"):
+            table = clause.table
 
+    if table is not None and "bind_key" in table.metadata.info:
+        key = table.metadata.info["bind_key"]
         if key not in engines:
             raise sa_exc.UnboundExecutionError(
                 f"Bind key '{key}' is not in 'SQLALCHEMY_BINDS' config."
--- a/src/flask_sqlalchemy/session.py
+++ b/src/flask_sqlalchemy/session.py
@@ -79,21 +79,23 @@ class Session(sa_orm.Session):
 
 
 def _clause_to_engine(
-    clause: t.Any | None,
+    clause: sa.ClauseElement | None,
     engines: t.Mapping[str | None, sa.engine.Engine],
 ) -> sa.engine.Engine | None:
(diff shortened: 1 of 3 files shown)


src/flask_sqlalchemy/session.py
```diff
@@ -79,14 +79,21 @@ def get_bind(
...
+    engines: t.Mapping[str | None, sa.engine.Engine],
 ) -> sa.engine.Engine | None:
     """If the clause is a table, return the engine associated with the table's
     metadata's bind key.
     """
-    if isinstance(clause, sa.Table) and "bind_key" in clause.metadata.info:
-        key = clause.metadata.info["bind_key"]
-
+    table = None
+    if clause is not None:
```
davidism, line 89: Add an empty line between the previous code and a block start line.

src/flask_sqlalchemy/session.py
```diff
@@ -79,14 +79,21 @@ def get_bind(
...
+        if isinstance(clause, sa.Table):
+            table = clause
+        elif hasattr(clause, "table"):
+            table = clause.table
+
+    if table is not None and "bind_key" in table.metadata.info:
+        key = table.metadata.info["bind_key"]
         if key not in engines:
```
davidism, line 97: Add an empty line between the previous code and a block start line.

pamelafox:
@davidism I added the requested newlines. I also narrowed down the types and used isinstance checks to confirm the type of class containing the table. That was after digging through the SA code to determine where the "table" attribute came from. That might mean this fix won't work for other situations, but I haven't found other classes with a table attribute that seem like they'd come into play here. I also parameterized the test to confirm it works with 2.0 style.

PR #1218, opened 2023-06-09
use pip-compile-multi and flit_core instead of pdm

davidism:
I used this project to test switching to PDM to manage the development environment, lock dependencies, and run commands. However, it wasn't significantly better for our use case over using pip-compile-multi to manage `requirements/*.in` to `*.txt` files. Instead of going back to setuptools, I've used flit_core as the build backend, as it's simpler and faster.

This also enables PyPI's trusted publisher authentication.


davidism:
If you already have a dev env using pdm, you'll want to delete any local `.venv` and `.tox` folders, then create a new venv:

```
$ rm -r .tox .venv
$ python3 -m venv .venv
$ . .venv/bin/activate
$ pip install -r requirements/dev.txt && pip install -e .
```

PR #1253, opened 2023-09-03
Update flaskr and todo to use SQLAlchemy 2.0 style

pamelafox:
This PR updates the Flaskr and todo sample apps to use 2.0 style models. All flaskr tests pass.

Checklist:

~- [ ] Add tests that demonstrate the correct behavior of the change. Tests should fail without the change.~
~- [ ] Add or update relevant docs, in the docs folder and in code.~
~- [ ] Add an entry in `CHANGES.rst` summarizing the change and linking to the issue.~
~- [ ] Add `.. versionchanged::` entries in any relevant code docs.~

### DIFF
--- a/examples/flaskr/flaskr/__init__.py
+++ b/examples/flaskr/flaskr/__init__.py
@@ -4,10 +4,16 @@ import click
 from flask import Flask
 from flask.cli import with_appcontext
 from flask_sqlalchemy import SQLAlchemy
+from sqlalchemy.orm import DeclarativeBase
 
-__version__ = (1, 0, 0, "dev")
+__version__ = (1, 1, 0, "dev")
 
-db = SQLAlchemy()
+
+class Base(DeclarativeBase):
+    pass
+
+
+db = SQLAlchemy(model_class=Base)
 
 
 def create_app(test_config=None):
--- a/examples/flaskr/flaskr/auth/models.py
+++ b/examples/flaskr/flaskr/auth/models.py
@@ -1,22 +1,29 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from sqlalchemy.orm import Mapped, mapped_column, relationship
 from werkzeug.security import check_password_hash
 from werkzeug.security import generate_password_hash
 
 from flaskr import db
 
+if TYPE_CHECKING:
+    from ..blog.models import Post
+
 
 class User(db.Model):
-    id = db.Column(db.Integer, primary_key=True)
-    username = db.Column(db.String, unique=True, nullable=False)
-    password_hash = db.Column(db.String, nullable=False)
+    id: Mapped[int] = mapped_column(primary_key=True)
+    username: Mapped[str] = mapped_column(unique=True)
+    password_hash: Mapped[str]
+    posts: Mapped[list[Post]] = relationship("Post", back_populates="author")
 
-    posts = db.relationship("Post", back_populates="author")
-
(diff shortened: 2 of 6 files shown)


examples/flaskr/flaskr/auth/models.py
```diff
@@ -1,15 +1,20 @@
+from typing import TYPE_CHECKING
+
+from sqlalchemy.orm import Mapped, mapped_column, relationship
 from werkzeug.security import check_password_hash
 from werkzeug.security import generate_password_hash
 
 from flaskr import db
 
+if TYPE_CHECKING:
+    from flaskr.blog.models import Post
```
pamelafox, line 10: Needed due to circular import

examples/flaskr/flaskr/__init__.py
```diff
@@ -4,10 +4,16 @@
 from flask import Flask
 from flask.cli import with_appcontext
 from flask_sqlalchemy import SQLAlchemy
+from sqlalchemy.orm import DeclarativeBase
 
 __version__ = (1, 0, 0, "dev")
 
-db = SQLAlchemy()
+
+class Base(DeclarativeBase):
```
ThiefMaster, line 12: I don't think this is needed at all. `db = SQLAlchemy()` works fine even for SA 2, no?

If this is actually needed to use the `Mapped` notation this feels like something that needs to be fixed in FSA - because having to do this here is super ugly IMHO.

examples/flaskr/flaskr/auth/models.py
```diff
@@ -1,15 +1,20 @@
...
-class User(db.Model):
-    id = db.Column(db.Integer, primary_key=True)
-    username = db.Column(db.String, unique=True, nullable=False)
-    password_hash = db.Column(db.String, nullable=False)
 
-    posts = db.relationship("Post", back_populates="author")
+class User(db.Model):
+    id: Mapped[int] = mapped_column(primary_key=True)
+    username: Mapped[str] = mapped_column(unique=True, nullable=False)
+    password_hash: Mapped[str] = mapped_column(nullable=False)
```
ThiefMaster, line 16: ugh i can't believe that SA didn't add something nicer for the most likely very common case of "primitive type, but not nullable"...

examples/flaskr/flaskr/__init__.py
pamelafox: I'm not sure if you saw the discussion in https://github.com/pallets-eco/flask-sqlalchemy/issues/1140 but yes, the design we settled on does require you to pass in a base class. This gives you the flexibility to use MappedAsDataclass, DeclarativeBaseNoMeta, etc, instead of us deciding on what's in your base class.

ThiefMaster: nope, hadn't seen it or at least I don't remember it... feels a bit obscure when you just have a single superclass but makes more sense if you use multiple superclasses/mixins

davidism: We'd like to make the 2.0 style the default at some point, but it will require a bit more thought, and we didn't want to block being able to use it at all in this next release.

examples/flaskr/flaskr/auth/models.py
davidism: `Mapped[str]` does mean `nullable=False`, and `Mapped[str | None]` means `nullable=True`, so we can remove the `mapped_column` for some of these.

davidism (approved):

PR #1254, opened 2023-09-10
release version 3.1.0

pamelafox:


davidism:
We've kept `get_engine` until 3.2 to make the transition for Flask-Migrate users a little smoother. See https://github.com/miguelgrinberg/Flask-Migrate/pull/526 as well.

PR #1283, opened 2023-11-25
Fixing missing base class in reflection docs, adding to reflection test

### LINKED ISSUE #1280: Incorrect(?) example in documentation
Environment:

- Python version: 3.x
- Flask-SQLAlchemy version: 3.1.x
- SQLAlchemy version: N/A
I found a possibly incorrect example in the Reflecting Tables section of docs/models.rst
There is a class definistion:
```

class User:
    __table__ = db.metadatas["auth"].tables["user"]
```

As far as I know it doesn't work, because the parent class is missing from it.
The correct definition would be

```
class User(Base):
    __table__ = db.metadatas["auth"].tables["user"]

```
Of course, I may have misunderstood something.

pamelafox:
This PR adds "(db.Model)" to reflection docs example. I also added to the reflection test to mirror the documentation example, and I put an additional example in the docs showing default bind key. 

- fixes #1280

Checklist:

- [X] Add or update relevant docs, in the docs folder and in code.
~- [ ] Add an entry in `CHANGES.rst` summarizing the change and linking to the issue.~
~- [ ] Add `.. versionchanged::` entries in any relevant code docs.~

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -37,11 +37,5 @@ jobs:
           python-version: ${{ matrix.python }}
           cache: 'pip'
           cache-dependency-path: 'requirements/*.txt'
-      - name: cache mypy
-        uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8
-        with:
-          path: ./.mypy_cache
-          key: mypy|${{ matrix.python }}|${{ hashFiles('requirements/mypy.txt') }}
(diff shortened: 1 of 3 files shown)


pamelafox:
The typing test is now passing, though I'm confused as to why it was previously failing. Seems like it must be cache related since I can't replicate locally, but I also haven't clearly proved it was cache related. Anyway, passing now.

davidism (approved):

PR #1308, opened 2024-01-31
Stop using datetime.utcnow() in tests

### LINKED ISSUE #1303: Python 3.12 changes cause test failures due to deprecation warning
Running the test suite under Python 3.12 has multiple failures that all stem from:

`DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).`

Since warnings are treated as errors, this causes a failure.

s-t-e-v-e-n-k:
datetime.utcnow() is deprecated for Python 3.12+, and raises a warning. Since warnings are treated as errors, this results in test failures. Since utcnow calls are done by the SQLAlchemy mapping machinery, we need to use a callable.

Fixes #1303

Checklist:

### DIFF
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import typing as t
 from datetime import datetime
+from datetime import timezone
 
 import pytest
 import sqlalchemy as sa
@@ -12,6 +13,10 @@ from flask import Flask
 from flask_sqlalchemy import SQLAlchemy
 from flask_sqlalchemy.model import DefaultMeta
 from flask_sqlalchemy.model import Model
+
+
+def utc_now() -> datetime:
+    return datetime.now(tz=timezone.utc)
 
 
 def test_default_model_class_1x(app: Flask) -> None:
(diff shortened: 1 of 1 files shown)


pamelafox (changes requested):
Thanks! As this is a bug fix, could you make it off the latest 3.x branch, per https://github.com/pallets-eco/flask-sqlalchemy/blob/main/CONTRIBUTING.rst#start-coding ?

tests/test_model.py
```diff
@@ -14,6 +15,11 @@
 from flask_sqlalchemy.model import Model
 
 
+class UTCNow(datetime):
```
davidism, line 18: Rather than making this a subclass with a tricky `__new__`, this should just be a function:

```python
def now() -> datetime:
    return datetime.now(timezone.utc)
```

s-t-e-v-e-n-k:
I have addressed both comments, thanks!

pamelafox (approved):

CHANGES.rst
```diff
@@ -2,6 +2,7 @@ Version 3.1.2
 -------------
 
 - Fix issue with calling ``repr()`` on ``SQLAlchemy`` instance with no default engine. :issue:`1295`
+-  No longer call ``datetime.utcnow()`` in the test suite. :issue:`1303`
```
davidism, line 5: Typically we don't include a changes entry for tests and docs changes, as they don't affect how the user uses the library. Sorry, missed this earlier.

s-t-e-v-e-n-k:
@davidism Thanks, I've reverted that bit.

davidism:
Sorry, completely forgot about this PR and ended up fixing it myself when doing some updates. Thanks for working on it!

PR #1315, opened 2024-02-18
Bring up URI check for sqlite engine

### LINKED ISSUE #1314: Using SQLite URI format for in-memory database still creates a file
## Outline

When using the SQLite URI format for an in memory database `sqlite:///file::memory:?uri=true` Flask-SQLAlchemy still creates a database file on disk.

## MRE

You can replicate this behaviour locally with the following code:

```python3
from flask import Flask
from sqlalchemy.orm import DeclarativeBase, mapped_column, Mapped

class Base(DeclarativeBase):
    pass

db = SQLAlchemy(model_class=Base)

class User(db.Model):
    id: Mapped[int] = mapped_column(primary_key=True)
    email: Mapped[str]

app = Flask(__name__)

app.config["SQLALCHEMY_DATABASE_URI"] = "sqlite:///file::memory:?uri=true"

db.init_app(app)

with app.app_context():
    db.create_all()

@app.route("/")
def index():
    return "<h1>Hello world</h1>"
```

If you start the app with `flask run --debug` you'll find you have an `instance/:memory:` file on disk.

## Expected behaviour

Running the above snippet does not produce a file on disk.

## Environment:

- Python version: Python 3.11.7
- SQLAlchemy version: 2.0.27

Sparrow0hawk:
## Changes

This PR brings up the check for a SQLite URI within the `_apply_driver_defaults` function to allow for earlier detection of if the database should run in-memory.

It adds a test to check that a database file is not created when using a SQLite URI filename for an in-memory database.

- fixes #1314 

Checklist:


src/flask_sqlalchemy/extension.py
```diff
@@ -607,22 +607,19 @@ def _apply_driver_defaults(self, options: dict[str, t.Any], app: Flask) -> None:
         url = sa.engine.make_url(options["url"])
 
         if url.drivername in {"sqlite", "sqlite+pysqlite"}:
-            if url.database is None or url.database in {"", ":memory:"}:
+            # the url might look like sqlite:///file:path?uri=true
+            is_uri = url.query.get("uri", False)
+
+            db_str = url.database[5:] if is_uri and url.database else url.database
```
davidism, line 613: Keep this as separate blocks, avoid ternary unless the line stays very short.

PR #1349, opened 2024-06-24
Improve the example in README.md

yuxiaoy1:
There are three updates:
- drop the table before create it, to fix the `UNIQUE constraint failed` error  when re-run the example code.
- change username `example` to `frank` to fit better with person's name.
- use shorthand `scalars()` method and print all usernames.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,6 @@ easier to accomplish common tasks.
 [Flask]: https://flask.palletsprojects.com
 [SQLAlchemy]: https://www.sqlalchemy.org
 
-
 ## Pallets Community Ecosystem
 
 > [!IMPORTANT]\
@@ -18,7 +17,6 @@ easier to accomplish common tasks.
 > this project, please reach out on [the Pallets Discord server][discord].
 >
 > [discord]: https://discord.gg/pallets
-
 
 ## A Simple Example
 
@@ -45,5 +43,5 @@ with app.app_context():
     db.session.add(User(username="example"))
     db.session.commit()
 
-    users = db.session.execute(db.select(User)).scalars()
+    users = db.session.scalars(db.select(User))
 ```


README.md
```diff
@@ -40,10 +40,12 @@ class User(db.Model):
...
+    db.drop_all()
     db.create_all()
 
-    db.session.add(User(username="example"))
+    db.session.add(User(username="frank"))
     db.session.commit()
 
-    users = db.session.execute(db.select(User)).scalars()
+    users = db.session.scalars(db.select(User))
+    print({"users": [u.username for u in users]})
```
pamelafox, line 50: I don't love over-encouraging print statements, but it is common in examples, so I see the reasoning.

I think this could at least be simplified, as the dict seems unneeded:

```suggestion
    print([u.username for u in users])
```

If you extend from MappedAsDataClass, then you can just print(users) but you don't get that nice repr with just the DeclarativeBase, so the list comprehension is warranted.

README.md
```diff
@@ -40,10 +40,12 @@ class User(db.Model):
     username: Mapped[str] = mapped_column(unique=True)
 
 with app.app_context():
+    db.drop_all()
```
pamelafox, line 43: @davidism I see that https://pypi.org/project/Flask-SQLAlchemy-Lite/ has the same example, so what do you think of the changes, as you'd probably want them there too? I do use drop_all in my non-Flask SQLAlchemy examples, as it does make them easier to re-run, there's just a slight risk of someone running that in production. But hopefully a very small risk as they need to change all this code.

README.md
yuxiaoy1: Thanks for the review, but I see the dict a more friendly output here.

davidism:
I'm fine with the `scalars` change. I think that needs to be applied throught the docs though, I wasn't familiar with all the new patterns yet when the docs were cleaned up. Everything else, while it makes sense, I don't think needs to be in the example.

README.md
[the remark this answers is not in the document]
yuxiaoy1: It's being used in flaskr example, there might be updates to that as well.

pamelafox (approved):
Thanks, approving and merging.
Note that using execute isn't stringly wrong, per https://docs.sqlalchemy.org/en/20/changelog/migration_20.html
but this is shorter.

PR #1390, opened 2025-06-01
default_per_page in Pagination

### LINKED ISSUE #1362: Limit = 0 provides a configurable per_page
In this code https://github.com/pallets-eco/flask-sqlalchemy/blob/main/src/flask_sqlalchemy/pagination.py#L132

When per_page = 0, it will be set to 20 if error_out = False. 

```
        if per_page < 1:
            if error_out:
                abort(404)
            else:
                per_page = 20
```

20 might not be the case for all scenarioes, better give it a configurable value.

py39cptCiolacu:
I've added a new attribute for Pagination - default_per_page.

- if per_page is not provided per_page = default_per_page
- if per_page is invalid (less than 1) AND error_out = False, per_page = default_per_page

fixes #1362


src/flask_sqlalchemy/pagination.py
```diff
@@ -92,6 +97,7 @@ def _prepare_page_args(
         *,
         page: int | None = None,
         per_page: int | None = None,
+        default_per_page: int | None = None,
```
davidism: The code here should be using `self.default_per_page`, not an argument.

src/flask_sqlalchemy/pagination.py
```diff
@@ -47,6 +47,7 @@ def __init__(
         self,
         page: int | None = None,
         per_page: int | None = None,
+        default_per_page: int | None = None,
```
davidism: Use 20 as the default, it shouldn't be optional.

omoh09:
Hi @davidism , I noticed this PR adds default_per_page. I’d be happy to help fix the failing tests or docs if that’s still needed, could you confirm?

PR #1405, opened 2025-10-31
Fix deprecation warning in `Query.get_or_404()` for SQLAlchemy 2.x

### LINKED ISSUE #1404: `Query.get_or_404()` triggers deprecation warning under SQLAlchemy 2.x
### **Description**

When using Flask-SQLAlchemy 3.x with SQLAlchemy 2.x, calling `Query.get_or_404()` produces a deprecation warning indicating that `Query.get()` is considered legacy. The warning points to the internal use of `Query.get()` within `get_or_404()`.

### **Steps to Reproduce**

1. Use SQLAlchemy 2.x and Flask-SQLAlchemy 3.x.
2. Define a simple model and call:

   ```python
   user = User.query.get_or_404(1)
   ```
3. Observe the warning in the console or test output.

### **Observed Warning**

```
LegacyAPIWarning: The Query.get() method is considered legacy as of the 1.x series of SQLAlchemy 
and becomes a legacy construct in 2.0. The method is now available as Session.get().
(Background on SQLAlchemy 2.0 at: https://sqlalche.me/e/b8d9)
```

### **Expected Behavior**

`get_or_404()` should not emit any deprecation warnings when running under SQLAlchemy 2.x.

### **Environment**

- Python version: 3.12
- SQLAlchemy version: 2.0.44

mreid-tt:
### **Description**

This pull request updates `Query.get_or_404()` to remove the use of the deprecated `Query.get()` method, which triggers a `LegacyAPIWarning` under SQLAlchemy 2.x.

The change aligns `get_or_404()` with the SQLAlchemy 2.x API by using the session’s `get()` method instead. This preserves existing behavior while ensuring forward compatibility and eliminating the deprecation warning.

fixes #1404

### **Additional Notes**

* No behavioral changes are introduced beyond internal API alignment.
* `first_or_404()` and `one_or_404()` are unaffected, as they do not use deprecated methods.

### DIFF
--- a/src/flask_sqlalchemy/query.py
+++ b/src/flask_sqlalchemy/query.py
@@ -27,7 +27,8 @@ class Query(sa_orm.Query):  # type: ignore[type-arg]
         :param ident: The primary key to query.
         :param description: A custom message to show on the error page.
         """
-        rv = self.get(ident)
+        mapper = self._only_full_mapper_zero("get")
+        rv = self.session.get(mapper.class_, ident)
 
         if rv is None:
             abort(404, description=description)
--- a/src/flask_sqlalchemy/query.py
+++ b/src/flask_sqlalchemy/query.py
@@ -27,8 +27,9 @@ class Query(sa_orm.Query):  # type: ignore[type-arg]
         :param ident: The primary key to query.
         :param description: A custom message to show on the error page.
         """
-        mapper = self._only_full_mapper_zero("get")
-        rv = self.session.get(mapper.class_, ident)
+        model = (self.column_descriptions[0].get("entity")
+            or self.column_descriptions[0]["type"])
+        rv = self.session.get(model, ident)
 
         if rv is None:
             abort(404, description=description)


src/flask_sqlalchemy/query.py
```diff
@@ -27,7 +27,8 @@ def get_or_404(self, ident: t.Any, description: str | None = None) -> t.Any:
         :param ident: The primary key to query.
         :param description: A custom message to show on the error page.
         """
-        rv = self.get(ident)
+        mapper = self._only_full_mapper_zero("get")
```
davidism: We should avoid private APIs. Make a request to SQLAlchemy to make this a publicly supported API first.

mreid-tt: Thanks for catching that! I’ve updated the implementation to use the public API instead of the private one. It now relies on column_descriptions to resolve the mapped class.

davidism:
Wait a second, this issue and PR make no sense. This still is a method on `Query`, which _is_ the legacy API. Using the new `select` API from within it makes no sense, just don't use `query` to begin with in that case.

I'd really appreciate if you don't use AI to continue this conversation.

mreid-tt:
> Wait a second, this issue and PR make no sense. This still is a method on `Query`, which _is_ the legacy API. Using the new `select` API from within it makes no sense, just don't use `query` to begin with in that case.

To be honest, I’m not a Python expert; I was mainly frustrated by the repeated deprecation warnings. Since the repository hasn’t seen much recent maintenance, I decided to try a small fix. The goal wasn’t a full refactor, just a minimal change to remove the warnings and hopefully support a minor release, given that it’s been about two years since the last one.

davidism:
Going to close this then. I'd recommend using flask-sqlalchemy-lite and moving to use the sqlalchemy select api directly, if all you want to do is use modern sqlalchemy without warnings.

davidism:
Also, we've already provided modern methods for quite some time: https://flask-sqlalchemy.readthedocs.io/en/stable/queries/#queries-for-views.

PR #1424, opened 2026-04-18
bug fix select pagination for multi-column results

### LINKED ISSUE #1423: paginate with join failed
# Environment:

- Python version: >=3.14
- SQLAlchemy version: 2.0.49

# how to reproduce
```python
from flask import Flask, jsonify
from sqlalchemy.orm import DeclarativeBase, Mapped, mapped_column
from sqlalchemy import select as sql_select
from dataclasses import dataclass

class Base(DeclarativeBase):
  pass

class User(Base):
    __tablename__ = "user"
    id: Mapped[int] = mapped_column(primary_key=True)
    email: Mapped[str]

class Other(Base):
    __tablename__ = "other"
    id: Mapped[int] = mapped_column(primary_key=True)
    age: Mapped[int]

db = SQLAlchemy(model_class=Base)

app = Flask(__name__)
# configure the SQLite database, relative to the app instance folder
app.config["SQLALCHEMY_DATABASE_URI"] = "sqlite:///project.db"
# initialize the app with the extension
db.init_app(app)

with app.app_context():
    db.drop_all()
    db.create_all()
    db.session.add(User(username="test", email="test@test.com"))
    db.session.add(Other(username="test", age=2026))
    db.session.commit()

@dataclass
class UserInfo:
    username: str
    email: str
    age: int

@app.route("/users")
def user_list():
    stmt = sql_select(User, Other).join(Other, User.username==Other.username).order_by(User.id)
    paginate = db.paginate(stmt)

    users = []
    for user, other in paginate.items:
        users.append(UserInfo(username=user.username, email=user.email, age=other.age))

    result = {
        "has_next": paginate.has_next,
        "users": users
    }
    return jsonify(result)

if __name__ == "__main__":
    app.run(debug=True)
```

got unexpected error

Image

# why failed

https://github.com/pallets-eco/flask-sqlalchemy/blob/168cb4b7b50fe5176307a10d873781bfafc6eeda/src/flask_sqlalchemy/pagination.py#L335-L339

`.scalars()` here causes.

# how to fix
```python
    def _query_items(self) -> list[t.Any]:
        select = self._query_args["select"]
        select = select.limit(self.per_page).offset(self._query_offset)
        session = self._query_args["session"]
        return list(session.execute(select).unique().all())
```

hxzhao527:
Return full result rows from SelectPagination instead of scalar values so paginated select queries preserve all selected columns.

fix #1423

### DIFF
--- a/src/flask_sqlalchemy/pagination.py
+++ b/src/flask_sqlalchemy/pagination.py
@@ -336,7 +336,7 @@ class SelectPagination(Pagination):
         select = self._query_args["select"]
         select = select.limit(self.per_page).offset(self._query_offset)
         session = self._query_args["session"]
-        return list(session.execute(select).unique().scalars())
+        return list(session.execute(select).unique().all())
 
     def _query_count(self) -> int:
         select = self._query_args["select"]


davidism:
This breaks the much more common case of selecting a single model and getting back a list of instances instead of tuples. It needs to support both, which is the problem, since that's no longer possible to detect with public SQLAlchemy APIs.

